PR #40, opened 2021-03-08
Fix for rpc error: code = Unknown desc = json: cannot unmarshal number 200000000000 into Go struct field .pairs.limits.maximal of type int

brandonshope:
This change fixes the json error in the REST API call - /v1/serviceinfo

rpc error: code = Unknown desc = json: cannot unmarshal number 200000000000 into Go struct field .pairs.limits.maximal of type int

admin@raspberrypi:~/.boltz-lnd/macaroons $ curl --insecure --header "Grpc-Metadata-macaroon: $(xxd -ps -u -c 1000 ~/.boltz-lnd/macaroons/admin.macaroon)" https://localhost:9003/v1/serviceinfo
{"code":2,"message":"json: cannot unmarshal number 200000000000 into Go struct field .pairs.limits.maximal of type int","details":[]}


brandonshope:
Seems several of these ( int 'json:""' ) need to be changed to int64. Without cross referencing all of them, not sure how many.

michael1011:
@brandonshope could you check my PR #41 and test if it works for you? I removed pretty much all usages of `int` to avoid all potential issues with this cause down the road

brandonshope:
@michael1011 tested PR #41 on my node and it works fine. Thanks for all the work! Cheers!

PR #51, opened 2021-12-07
Build docker image for amd64, arm64, and arm

aphex3k:
This will enable usage of the docker image on raspberry pi and Apple M1 platform

image

See: https://hub.docker.com/r/aphex3k/boltz-lnd/tags

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,6 @@ changelog:
 
 docker:
 	@$(call print, "Building docker image")
-	docker build -t boltz/boltz-lnd:$(version) .
+	buildx build --platform linux/amd64 --platform linux/arm64 --platform linux/arm/v7 -t boltz/boltz-lnd:$(version) -t boltz/boltz-lnd:latest .
 
 .PHONY: build binaries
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -96,6 +96,6 @@ changelog:
 
 docker:
 	@$(call print, "Building docker image")
-	docker build -t boltz/boltz-lnd:$(version) .
+	docker buildx build --platform linux/amd64 --platform linux/arm64 --platform linux/arm/v7 -t boltz/boltz-lnd:$(version) -t boltz/boltz-lnd:latest .
```
michael1011, line 99: Which tags do you need for the docker image? Versions for the releases or `latest`?

michael1011 (commented):
Sorry for the inactivity. Thank you for the PR!

Just one little question

Makefile
aphex3k: Ideally you would have both tags. That would allow for pinning applications to a specific tagged version as well as "automatically" updating to the latest version available when using the `latest` tag.

As an example, look at the many tags provide by [node on dockerhub](https://hub.docker.com/_/node). While this is certainly overkill, it shows the benefits of providing multiple tags for each image.

michael1011:
Pushed the images!

PR #57, opened 2023-04-08
feat: allow setting claim transaction fee

michael1011:


armurbalda:
We should also communicate this to RTL once available in umbrel

Thompson1985 (approved):

PR #73, opened 2024-01-28
refactor: allow specifying multiple channel ids instead of one

jackstar12:
closes: #43


cln/cln.go
```diff
@@ -192,18 +193,18 @@ func (c *Cln) CreateInvoice(value int64, preimage []byte, expiry int64, memo str
...
 }
 
-func (c *Cln) PayInvoice(invoice string, feeLimit uint, timeoutSeconds uint, chanId lightning.ChanId) (*lightning.PayInvoiceResponse, error) {
+func (c *Cln) PayInvoice(invoice string, feeLimit uint, timeoutSeconds uint, chanIds []lightning.ChanId) (*lightning.PayInvoiceResponse, error) {
 	retry := uint32(timeoutSeconds)
 
 	var exclude []string
 
-	if chanId != 0 {
+	if len(chanIds) > 0 && chanIds[0] != 0 {
```
michael1011, line 201:
```suggestion
	if len(chanIds) > 0 {
```
Why do we still need this check?

lnd/lnd.go
```diff
@@ -282,28 +282,23 @@ func (lnd *LND) GetChannelInfo(chanId uint64) (*lnrpc.ChannelEdge, error) {
...
+func (lnd *LND) PayInvoice(invoice string, feeLimit uint, timeoutSeconds uint, chanIds []lightning.ChanId) (*lightning.PayInvoiceResponse, error) {
+	var outgoungIds []uint64
+	for _, chanId := range chanIds {
+		if chanId != 0 {
+			outgoungIds = append(outgoungIds, uint64(chanId))
 		}
-	*/
+	}
+
+	fmt.Println(outgoungIds)
```
michael1011, line 293: Rm

lnd/lnd.go
```diff
@@ -282,28 +282,23 @@ func (lnd *LND) GetChannelInfo(chanId uint64) (*lnrpc.ChannelEdge, error) {
...
 
 	client, err := lnd.router.SendPaymentV2(lnd.ctx, &routerrpc.SendPaymentRequest{
 		PaymentRequest:    invoice,
 		TimeoutSeconds:    int32(timeoutSeconds),
 		FeeLimitSat:       int64(feeLimit),
 		NoInflightUpdates: true,
 		// TODO: this is not working right now
```
michael1011, line 300: This you manage to reliable reproduce the flakiness?

michael1011 (changes requested):

cln/cln.go
jackstar12: Autoswap recommendations include a channel id, which is passed to the create swap call, but the zero value means no channel. I guess that check could be put somewhere else

lnd/lnd.go
jackstar12: In regtest:
```bash
clnId=$(lightning-cli-sim 1 getinfo | jq -r  .id)
chanId=$(lncli-sim 1 listchannels --peer $clnId | jq -r  ".channels.[].chan_id")
./boltzcli createreverseswap --chan-id $chanId  100000
```

jackstar12: its a regtest issue, works in testnet

michael1011 (approved):
utACK

Issue #71, opened 2024-01-26
Switch to APIv2+Taproot Swaps


kilrau:
Waiting for Taproot regular claims and APIv2 to be final.

PR #83, opened 2024-02-06
feat: v2 api

jackstar12:
closes: #71


boltz/liquid.go
```diff
@@ -234,63 +246,50 @@ func constructLiquidTransaction(network *Network, outputs []OutputDetails, outpu
...
+			sigHash := liquidTaprootHash(tx, network, outputs, i, false)
+			signature, err := schnorr.Sign(output.PrivateKey, sigHash[:])
+			if err != nil {
 				return nil, err
 			}
 
-			valid, err := p.ValidateInputSignatures(i)
+			if !signature.Verify(sigHash[:], output.PrivateKey.PubKey()) {
+				return nil, errors.New("Signature verification failed")
+			}
```
michael1011, line 261: Why do we verify our own signature?

boltz/musig.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return err
+	}
+
+	nonce, err := hex.DecodeString(boltzSignature.PubNonce)
+	if err != nil {
+		return err
+	}
+
+	all, err := session.RegisterPubNonce([66]byte(nonce))
```
michael1011, line 51: A check to make sure the length of `nonce` is actually 66 before doing the cast would be appropriate

boltz/swaptree.go
```diff
@@ -0,0 +1,294 @@
...
+	expectedScript, err := expected.Script()
+	if err != nil {
+		return errors.New("expected script is invalid")
+	}
+
+	if !bytes.Equal(actual, expectedScript) {
+		expected, _ := txscript.DisasmString(expectedScript)
+		fmt.Printf("expected:\n%v\n", expected)
+		actual, _ := txscript.DisasmString(actual)
+		fmt.Printf("actual:\n%v\n", actual)
```
michael1011, line 244: rm prints

database/migration.go
```diff
@@ -175,122 +185,194 @@ func (database *Database) performMigration(oldVersion int) error {
...
+		`
+		if _, err := tx.Exec(migration); err != nil {
+			return err
+		}
 
-		return database.postMigration(oldVersion)
+		return tx.postMigration(oldVersion)
 	case latestSchemaVersion:
-		logger.Info("Database already at latest schema version: " + strconv.Itoa(latestSchemaVersion))
+		logger.Info("tx already at latest schema version: " + strconv.Itoa(latestSchemaVersion))
```
michael1011, line 375:
```suggestion
		logger.Info("Database already at latest schema version: " + strconv.Itoa(latestSchemaVersion))
```

nursery/reverse.go
```diff
@@ -142,22 +142,60 @@ func (nursery *Nursery) handleReverseSwapStatus(reverseSwap *database.ReverseSwa
...
 
+		lockupVout, lockupValue, err := lockupTx.FindVout(nursery.network, lockupAddress)
 		if err != nil {
-			handleError("Could not finalize claim transaction: " + err.Error())
+			handleError("Could not find lockup vout of Reverse Swap " + reverseSwap.Id)
 			return
 		}
 
+		if lockupValue < reverseSwap.OnchainAmount {
+			logger.Warn("Boltz locked up less onchain coins than expected. Abandoning Reverse Swap")
```
michael1011, line 172: We should also handle that and not just log

boltz/liquid.go
jackstar12: debugging leftovers

boltz/api.go
```diff
@@ -320,6 +413,62 @@ func (boltz *Boltz) CreateReverseSwap(request CreateReverseSwapRequest) (*Create
 	return &response, err
 }
 
+func (boltz *Boltz) ConstructTransaction(network *Network, currency Currency, outputs []OutputDetails, address string, feeSatPerVbyte float64) (Transaction, uint64, error) {
```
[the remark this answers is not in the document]
jackstar12: Im not sure either. But I also like it since the consumer of the function doesnt have to care about any musig stuff

boltz/btc.go
```diff
@@ -78,8 +110,10 @@ func constructBtcTransaction(network *Network, outputs []OutputDetails, outputAd
 
 	for _, output := range outputs {
 		// Set the highest timeout block height as locktime
-		if output.TimeoutBlockHeight > transaction.LockTime {
-			transaction.LockTime = output.TimeoutBlockHeight
+		if !output.Cooperative {
+			if output.TimeoutBlockHeight > transaction.LockTime {
+				transaction.LockTime = output.TimeoutBlockHeight
+			}
```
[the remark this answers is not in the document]
jackstar12: It has to be this way since cooperative refunds dont have a timelock

boltz/btc.go
```diff
@@ -49,23 +50,53 @@ func (transaction *BtcTransaction) VSize() uint64 {
...
-		if err != nil {
-			continue
-		}
-
-		for _, outputAddress := range outputAddresses {
-			if outputAddress.EncodeAddress() == addressToFind {
-				return uint32(vout), uint64(output.Value), nil
-			}
+		// first 2 bytes are witness type and length
+		if bytes.Equal(output.PkScript[2:], toFind.ScriptAddress()) {
```
michael1011, line 59: Add a check if the `output.PkScript` has more than 3 bytes. It's possible and just in case a tx like that is ever used with this function, it should not panic

database/migration.go
```diff
@@ -175,141 +185,205 @@ func (database *Database) performMigration(oldVersion int) error {
...
-		_, err = database.Exec("DROP TABLE channelCreations")
+	case 3:
+		logMigration(oldVersion)
+
+		rows, err := tx.Query("SELECT id FROM swaps WHERE state = ?", boltzrpc.SwapState_PENDING)
 		if err != nil {
 			return err
 		}
+		if rows.Next() {
+			return errors.New("tx migration failed: found pending swaps")
```
michael1011, line 299:
```suggestion
			return errors.New("database migration failed: found pending swaps")
```

database/migration.go
```diff
@@ -175,141 +185,205 @@ func (database *Database) performMigration(oldVersion int) error {
...
+		}
 
-		_, err = database.Exec("UPDATE version SET version = 3")
+		rows, err = tx.Query("SELECT id FROM reverseSwaps WHERE state = ?", boltzrpc.SwapState_PENDING)
 		if err != nil {
 			return err
 		}
+		if rows.Next() {
+			return errors.New("tx migration failed: found pending reverse swaps")
```
michael1011, line 307:
```suggestion
			return errors.New("database migration failed: found pending reverse swaps")
```

lnd/lnd.go
```diff
@@ -205,12 +204,16 @@ func (lnd *LND) ListChannels() ([]*lightning.LightningChannel, error) {
...
-	invoice, err := lnd.client.AddInvoice(lnd.ctx, &lnrpc.Invoice{
+	request := &lnrpc.Invoice{
 		Memo:      memo,
 		Value:     value,
 		Expiry:    expiry,
 		RPreimage: preimage,
-	})
+	}
+	if lnd.regtest {
+		request.CltvExpiry = 20
```
michael1011, line 214: This 20 magic value is used for LND and CLN. Let's make it a constant used by both
And add a comment why this is being done on regtest only

utils/connect.go
```diff
@@ -10,34 +10,43 @@ import (
...
+func ConnectBoltz(lightning lightning.LightningNode, boltz *boltz.Boltz) error {
+	nodes, err := boltz.GetNodes()
 	if err != nil {
-		return "", err
+		return err
 	}
 
 	symbol := "BTC"
-	node, hasNode := nodes.Nodes[symbol]
+	btcNodes, hasNode := nodes[symbol]
```
michael1011, line 38:
```suggestion
	nodesForSymbol, hasNodes := nodes[symbol]
```

utils/connect.go
```diff
@@ -10,34 +10,43 @@ import (
...
-	err = lightning.ConnectPeer(uri)
 
-	if err == nil {
-		logger.Info("Connected to Boltz node: " + uri)
-	} else if strings.Contains(err.Error(), "already connected to peer") {
-		logger.Info("Already connected to Boltz node: " + uri)
-		err = nil
-	} else {
-		err = fmt.Errorf("could not connect to boltz node %s: %w", uri, err)
+	if err := connect(lightning, btcNodes.CLN); err != nil {
```
michael1011, line 48: The `LND` and the `CLN` fields are both optional. Make sure there is no error thrown when the backend is running without one of those two

utils/connect.go
```diff
@@ -10,34 +10,43 @@ import (
 	"github.com/BoltzExchange/boltz-client/logger"
 )
 
-func ConnectBoltz(lightning lightning.LightningNode, boltz *boltz.Boltz) (string, error) {
-	nodes, err := boltz.GetNodes()
+func connect(node lightning.LightningNode, peer *boltz.NodeInfo) error {
+	if len(peer.Uris) == 0 {
+		return fmt.Errorf("no uris found for peer: %s", peer.PublicKey)
+	}
+	uri := peer.Uris[0]
```
michael1011, line 17: We should prefer clearnet URIs. The backend orders the URIs that way already, but just to the be sure, it should be done here too

PR #85, opened 2024-02-12
feat: replace sse with ws

jackstar12:


boltz/api.go
```diff
@@ -111,6 +111,7 @@ type NodeInfo struct {
 type Nodes = map[string]map[string]NodeInfo
 
 type SwapStatusResponse struct {
+	Id               string `json:"id"`
```
michael1011, line 114: To avoid trying to use the `id` in responses to GET requests don't add it here. But create a new types that extends this one that is only used in the websocket

boltz/ws.go
```diff
@@ -0,0 +1,161 @@
...
+		wsUrl.Scheme = "ws"
+	}
+
+	conn, _, err := websocket.DefaultDialer.Dial(wsUrl.String(), nil)
+	boltz.conn = conn
+	if err != nil {
+		return fmt.Errorf("could not connect to boltz ws at %s: %w", wsUrl, err)
+	}
+
+	logger.Infof("Connected to boltz ws at %s", wsUrl)
```
michael1011, line 71:
```suggestion
	logger.Infof("Connected to Boltz ws at %s", wsUrl)
```

boltz/ws.go
```diff
@@ -0,0 +1,161 @@
...
+	"errors"
+	"fmt"
+	"net/url"
+	"time"
+
+	"github.com/BoltzExchange/boltz-client/logger"
+	"github.com/gorilla/websocket"
+)
+
+const reconnectInterval = 15
```
michael1011, line 14:
```suggestion
const reconnectInterval = 15 * time.Second
```

boltz/ws.go
```diff
@@ -0,0 +1,161 @@
...
+					boltz.subscriptions <- true
+					continue
+				default:
+					logger.Warnf("unknown event: %s", response.Event)
+				}
+			}
+		}
+		for {
+			logger.Errorf("lost connection to boltz ws, reconnecting in %d seconds", reconnectInterval)
+			time.Sleep(reconnectInterval * time.Second)
```
michael1011, line 128:
```suggestion
			time.Sleep(reconnectInterval)
```

boltz/ws.go
```diff
@@ -0,0 +1,161 @@
...
+				case "subscribe":
+					boltz.subscriptions <- true
+					continue
+				default:
+					logger.Warnf("unknown event: %s", response.Event)
+				}
+			}
+		}
+		for {
+			logger.Errorf("lost connection to boltz ws, reconnecting in %d seconds", reconnectInterval)
```
michael1011, line 127:
```suggestion
			logger.Errorf("lost connection to boltz ws, reconnecting in %d", reconnectInterval)
```

boltz/ws.go
```diff
@@ -0,0 +1,161 @@
...
+		"op":      "subscribe",
+		"channel": "swap.update",
+		"args":    swapIds,
+	}); err != nil {
+		return err
+	}
+	select {
+	case <-boltz.subscriptions:
+		return nil
+	case <-time.After(time.Second):
```
michael1011, line 153: One seconds is a little aggressive. Let's make that larger

boltz/ws.go
```diff
@@ -0,0 +1,161 @@
...
+
+const reconnectInterval = 15
+
+type SwapUpdate = SwapStatusResponse
+
+type BoltzWebsocket struct {
+	apiUrl  string
+	Updates chan SwapUpdate
+
+	subscriptions chan bool
```
michael1011, line 22: Why?

boltz/ws.go
```diff
@@ -0,0 +1,161 @@
...
+					continue
+				}
+
+				switch response.Event {
+				case "update":
+					switch response.Channel {
+					case "swap.update":
+						var updates []SwapUpdate
+						data, _ := json.Marshal(response.Args)
+						if err := json.Unmarshal(data, &updates); err != nil {
```
michael1011, line 109: Cast instead of serialize and deserialize

rpcserver/router.go
```diff
@@ -205,24 +205,14 @@ func (server *routedBoltzServer) ListSwaps(_ context.Context, request *boltzrpc.
...
-	// Try to find a Reverse Swap with that ID
-	reverseSwap, err := server.database.QueryReverseSwap(request.Id)
-
-	if err == nil {
-		return &boltzrpc.GetSwapInfoResponse{
-			ReverseSwap: serializeReverseSwap(reverseSwap),
-		}, nil
+	swap, reverseSwap, err := server.database.QueryAnySwap(request.Id)
+	if err != nil {
+		return nil, handleError(errors.New("could not find Swap or Reverse Swap with ID " + request.Id))
```
michael1011, line 210:
```suggestion
		return nil, handleError(errors.New("could not find Swap with ID " + request.Id))
```

utils/channelforwarder.go
```diff
@@ -13,8 +15,9 @@ type ChannelForwarder[T any] struct {
 	lock        sync.Mutex
 }
 
-func ForwardChannel[T any](orig <-chan T, buffer int, saveValues bool) *ChannelForwarder[T] {
+func ForwardChannel[T any](orig chan T, buffer int, saveValues bool) *ChannelForwarder[T] {
 	cf := &ChannelForwarder[T]{
+		Original: orig,
```
michael1011, line 20: I'd prefer a method to write to the channel over exposing the channel variable

michael1011 (changes requested):

boltz/ws.go
jackstar12: I would keep it like that so I can use it in the log

boltz/ws.go
jackstar12: I cant type cast from because its initially parsed as a `map[string]any`

boltz/ws.go
jackstar12: It knows how to print it, nevermind

michael1011 (approved):

PR #89, opened 2024-02-28
feat: refund rpc

jackstar12:
- feat: `RefundSwap` rpc
- feat: database lock


cmd/boltzcli/commands.go
```diff
@@ -899,6 +899,22 @@ func createSwap(ctx *cli.Context) error {
...
+	ArgsUsage: "id addresss",
+	Action:    requireNArgs(2, refundSwap),
+}
+
+func refundSwap(ctx *cli.Context) error {
+	client := getClient(ctx)
+	id := ctx.Args().First()
+	address := ctx.Args().Get(1)
+	_, err := client.RefundSwap(id, address)
+	return err
```
michael1011, line 915: It would be nice to print the transaction id

PR #92, opened 2024-03-07
fix: proper semver version check

jackstar12:


cmd/boltzd/version.go
```diff
@@ -7,33 +7,32 @@ import (
...
 
 const minLndVersion = "0.15.0"
 const minClnVersion = "0.23.0"
-const minBoltzVersion = "2.3.0"
+const minBoltzVersion = "3.4.0"
 
 func checkVersion(name string, version string, minVersion string) {
-	versionInt, err := parseVersion(version)
-
+	parsed, err := semver.NewVersion(version)
```
michael1011, line 18: Some tests would be nice to ensure that we don't have regressions here

cmd/boltzd/version.go
```diff
@@ -7,33 +7,32 @@ import (
...
-	checkVersion("LND", info.Version, minLndVersion)
+	// remove the commit part from the version
+	version := strings.Split(info.Version, " ")[0]
+	checkVersion("LND", version, minLndVersion)
 }
 
 func checkClnVersion(info *lightning.LightningInfo) {
-	version, _ := strings.CutPrefix(info.Version, "v")
-	checkVersion("CLN", version, minClnVersion)
+	checkVersion("CLN", info.Version, minClnVersion)
```
michael1011, line 35: Test cases for CLN `v23.11.2-modded` and `v23.11rc3`

michael1011 (changes requested):

utils/version.go
```diff
@@ -0,0 +1,26 @@
...
+	// strip cln rc
+	version = strings.Split(version, "rc")[0]
+	version = strings.Split(version, "-")[0]
+
+	parsed, err := semver.NewVersion(version)
+	if err != nil {
+		return fmt.Errorf("Could not parse %s version %s: %s", name, version, err.Error())
+	}
+
+	if parsed.LessThan(semver.MustParse(minVersion)) {
```
michael1011, line 22: `MustParse` panics on error. Not a fan. Could we use the function that returns an error and handle it gracefully?

michael1011 (approved):
One minor thing

PR #93, opened 2024-03-07
refactor: use proper message type for autoswap config instead of json

jackstar12:


rpcserver/autoswap.go
```diff
@@ -119,20 +102,56 @@ func (server *routedAutoSwapServer) ReloadConfig(ctx context.Context, _ *empty.E
...
+//			FailureBackoff:      request.FailureBackoff,
+//			Budget:              request.Budget,
+//			BudgetInterval:      request.BudgetInterval,
+//			Currency:            boltz.ParseCurrency(request.Currency),
+//			Type:                request.Type,
+//			PerChannel:          request.PerChannel,
+//			Wallet:              request.Wallet,
+//			MaxSwapAmount:       request.MaxSwapAmount,
+//		}, nil
+//	}
```
michael1011, line 149: rm

autoswap/config.go
```diff
@@ -58,30 +66,22 @@ func (b Balance) String() string {
...
-	Type                boltz.SwapType
-	PerChannel          bool
-	Wallet              string
-	MaxSwapAmount       uint64
-
-	maxBalance   Balance
-	minBalance   Balance
-	strategy     Strategy
-	strategyName string
+	*SerializedConfig
```
michael1011, line 72: Wouldn't it be smarter to have this as a function so that in case a value is changed, we can call a function and get the latest version of the serialized version?

michael1011 (changes requested):

autoswap/config.go
jackstar12: You mean like this?
https://github.com/BoltzExchange/boltz-client/pull/93/files/4fa62f7756b63ad5b27c627d825c08061b260597#r1517672007
Well need all the fields in the go version aswell in that case

PR #94, opened 2024-03-08
feat: return_immediately parameter for reverse swap creation

jackstar12:


boltzrpc/boltzrpc.proto
```diff
@@ -410,14 +410,16 @@ message CreateReverseSwapRequest {
     Pair pair = 4;
     repeated string chan_ids = 5;
     optional string wallet = 6;
+    // The daemon will return immediately after zero conf swaps aswell.
```
michael1011, line 413: Didn't always return instantly for 0-conf? Could you elaborate a little bit about the different behaviours here?

boltzrpc/boltzrpc.proto
```diff
@@ -410,14 +410,16 @@ message CreateReverseSwapRequest {
...
 message CreateReverseSwapResponse {
     string id = 1;
     string lockup_address = 2;
-    uint32 routing_fee_milli_sat = 3 [deprecated = true];
 
-    // Only populated when 0-conf is accepted
-    string claim_transaction_id = 4 [deprecated = true];
+    // Only populated when 0-conf is accepted and return_immediately is set to false
+    uint64 routing_fee_milli_sat = 3;
+    string claim_transaction_id = 4;
```
michael1011, line 422: Make them optional when they are optional

michael1011, line 422: Also put the comment in this line so that it gets populated to the docs for both values

michael1011 (changes requested):

boltzrpc/boltzrpc.proto
jackstar12: Correct. And if this isnt set to true it will keep doing it, so its backwards compatible.
Clients who implement the update stream dont want that blocking call so they set it to true

boltzrpc/boltzrpc.proto
jackstar12: i think this would be a breaking change

PR #102, opened 2024-03-13
feat: noTls option and cors for rpcserver

jackstar12:
Makes it easier to use the rest api from other applications or browsers

### DIFF
--- a/boltzrpc/client/connection.go
+++ b/boltzrpc/client/connection.go
@@ -5,11 +5,13 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials"
-	"google.golang.org/grpc/metadata"
 	"os"
 	"strconv"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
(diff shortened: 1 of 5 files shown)


boltzrpc/client/connection.go
```diff
@@ -26,11 +28,16 @@ type Connection struct {
 }
 
 func (connection *Connection) Connect() error {
-	creds, err := credentials.NewClientTLSFromFile(connection.TlsCertPath, "")
+	creds := insecure.NewCredentials()
+	var err error
```
michael1011, line 32: Why is `err` declared here?

rpcserver/server.go
```diff
@@ -117,7 +120,15 @@ func (server *RpcServer) Init(
 	swapper.Init(database, onchain, autoSwapConfigPath)
 	routedServer.swapper = swapper
 
-	if !server.NoTls {
+	if server.NoTls {
+		// cleanup previous certificates to avoid confusions
```
michael1011, line 124:
```suggestion
		// cleanup previous certificates to avoid confusion
```

michael1011 (approved):

boltzrpc/client/connection.go
jackstar12: So that we can do a `=` assignment to the `creds` var from line 31.

PR #124, opened 2024-03-25
feat: global swap info stream

jackstar12:
add global events stream very useful for lnbits / btcpay fundingsources


dni (approved):
tested with boltz lnbits funding source

michael1011 (approved):

PR #130, opened 2024-03-26
refactor: per-output addresses for transactions

jackstar12:
Allow specifying a destination address for each output when creating a transaction. This simplifies any form of batching, currently swap refunds, but can be used for deferred claims of reverse / chain swaps aswell

### DIFF
--- a/boltz/liquid.go
+++ b/boltz/liquid.go
@@ -175,7 +175,10 @@ func constructLiquidTransaction(network *Network, outputs []OutputDetails, fee u
 			Amount: fee,
 		},
 	}
-	feePerOutput := fee / uint64(len(outValues))
+
+	outLen := uint64(len(outValues))
+	feePerOutput := fee / outLen
+	feeRemainder := fee % outLen
 	var blindingKeyCompressed []byte
 	var blinderIndex uint32
 	for rawAddres, value := range outValues {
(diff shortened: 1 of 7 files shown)


boltz/liquid.go
```diff
@@ -182,26 +159,59 @@ func constructLiquidTransaction(network *Network, outputs []OutputDetails, outpu
...
 
-	if err := updater.AddOutputs([]psetv2.OutputArgs{
-		{
-			Asset:        btcAsset,
-			Amount:       inputSum - fee,
-			Script:       script,
-			BlindingKey:  blindingKeyCompressed,
-			BlinderIndex: 0,
-		},
+	var blindingKeyCompressed []byte
```
michael1011, line 172: Should be defined in the loop where it is being used

boltz/liquid.go
```diff
@@ -182,26 +159,59 @@ func constructLiquidTransaction(network *Network, outputs []OutputDetails, outpu
...
+	txOutputs := []psetv2.OutputArgs{
 		{
 			Asset:  btcAsset,
 			Amount: fee,
 		},
-	}); err != nil {
+	}
+
+	feePerOutput := fee / uint64(len(outValues))
+	blinderIndex := 0
```
michael1011, line 181:
```suggestion
	var blinderIndex uint32 = 0
```

boltz/liquid.go
```diff
@@ -182,26 +159,59 @@ func constructLiquidTransaction(network *Network, outputs []OutputDetails, outpu
...
+		if err != nil {
+			return nil, errors.New("Could not generate output script: " + err.Error())
+		}
+
+		txOutputs = append(txOutputs, psetv2.OutputArgs{
+			Asset:        btcAsset,
+			Amount:       value - feePerOutput,
+			Script:       script,
+			BlindingKey:  blindingKeyCompressed,
+			BlinderIndex: uint32(blinderIndex),
```
michael1011, line 209:
```suggestion
			BlinderIndex: blinderIndex,
```

boltz/liquid.go
jackstar12: Its required after the loop aswell

PR #136, opened 2024-04-18
feat: chain swaps cli

jackstar12:


kilrau (changes requested):
## `createswap` & `createchainswap` parity:
- [x] to be in line with behavior of `createchainswap`, `createswap` should also print help output instead of starting an "any amount" swap process when called without args. It even says `--any-amount` is `(default: false)` but well that's not true
- [x] `boltzcli createswap --any-amount` defaults to mainchain, the `--currency` arg should NOT be optional and print help output. If you absolutely want to default, default to Liquid and remove the `--liquid` flag.
- [ ] I don't get why `createswap` has `--internal-send` AND `--wallet`. `--internal-send` sounds very vague and confusing so let's get rid of it and stick with `--from-wallet` & `--to-wallet` everywhere, just like in `createchainswap`. It's desirable that the user specifies which wallet to use and not automagically choose.
- [x] `createchainswap` should have the `--any-amount` feature too
- [x] There should be examples in both help outputs for `--any-amount`
- [ ] I like `--to-external` better to be in line with `--from-external`. Help output tells me that one has to be a currency, one an address

Adding more later

jackstar12:
> [ ] to be in line with behavior of `createchainswap`, `createswap` should also print help output instead of starting an "any amount" swap process when called without args. It even says `--any-amount` is `(default: false)` but well that's not true

This isnt supported by the backend yet.

> [ ] I don't get why `createswap` has `--internal-send` AND `--wallet`. `--internal-send` sounds very vague and confusing so let's get rid of it and stick with `--from-wallet` & `--to-wallet` everywhere, just like in `createchainswap`. It's desirable that the user specifies which wallet to use and not automagically choose.

`internal-send` is currently useful for clients who dont care about  any wallet stuff, e.g. RTL. 
But it should be  `external-pay`  in the cli for consistency.

kilrau (approved):
Looking good

boltz/liquid.go
```diff
@@ -70,6 +70,14 @@ func (transaction *LiquidTransaction) FindVout(network *Network, addressToFind s
 
 }
 
+func (transaction *LiquidTransaction) VoutValue(vout uint32) (uint64, error) {
+	result, err := confidential.UnblindOutputWithKey(transaction.Outputs[vout], transaction.OurOutputBlindingKey.Serialize())
+	if err != nil {
+		return 0, err
+	}
+	return result.Value, nil
```
michael1011, line 78: This assumes that we always blind. Which is the case right now, but in the future it would be nice to not throw in case we don't
That adds some complexity in the construction of the transaction, so let's leave that for later

cmd/boltzcli/commands.go
```diff
@@ -320,6 +330,29 @@ func swapInfoStream(ctx *cli.Context, id string, json bool) error {
...
+				}
+
+				status := boltz.ParseEvent(swap.Status)
+				switch status {
+				case boltz.SwapCreated:
+					fmt.Printf("Swap ID: %s\n", swap.Id)
+				case boltz.TransactionMempool:
+					fmt.Printf("User Transaction ID: %s\nAmount: %dsat\n", swap.FromData.GetLockupTransactionId(), swap.FromData.Amount)
+				case boltz.TransactionServerMempoool:
+					fmt.Printf("Server Transaction ID: %s\nAmount: %dsat\n", swap.ToData.GetLockupTransactionId(), swap.ToData.Amount)
```
michael1011, line 351: Maybe include the chain of the transaction in the message. Makes it easier to find. And lowercase transaction

michael1011 (approved):

PR #142, opened 2024-04-27
fix: disable rbf when sending from internal wallets

jackstar12:


onchain/wallet/wallet.go
```diff
@@ -572,12 +572,19 @@ func (wallet *Wallet) SendToAddress(address string, amount uint64, satPerVbyte f
...
 
 	var outputs struct {
-		Unspent any `json:"unspent_outputs"`
+		Unspent map[string][]map[string]any `json:"unspent_outputs"`
 	}
 	if err := withAuthHandler(C.GA_get_unspent_outputs(wallet.session, params, &handler), handler, &outputs); err != nil {
 		return "", err
 	}
 
+	// disable RBF
```
michael1011, line 581:
```suggestion
	// Disable RBF
```

michael1011 (changes requested):
What would be nice is an integration test that sends a transaction and checks that all sequences are set correctly. To make sure we don't have any regressions when updating GDK

michael1011 (approved):

PR #146, opened 2024-05-03
feat: magic routing hint support

jackstar12:


rpcserver/router.go
```diff
@@ -449,90 +492,99 @@ func (server *routedBoltzServer) createSwap(ctx context.Context, isAuto bool, re
 		return nil, handleError(err)
 	}
 
-	response, err := server.boltz.CreateSwap(createSwap)
+	if swapResponse == nil {
```
michael1011, line 495: Not a huge fan of this. Could we split the logic of this function into multiple ones for the different use cases it has now?

boltz/api.go
```diff
@@ -619,6 +621,24 @@ func (boltz *Boltz) RefundChainSwap(swapId string, request *RefundRequest) (*Par
...
+
+	Error string `json:"error"`
+}
+
+func (boltz *Boltz) GetReverseSwapBip21(invoice string) (*ReverseBip21, error) {
+	var response ReverseBip21
+	err := boltz.sendGetRequest(fmt.Sprintf("/v2/swap/reverse/%s/bip21", invoice), &response)
+
+	if response.Error != "" {
+		return nil, Error(errors.New(response.Error))
```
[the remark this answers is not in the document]
jackstar12: These are wrapped so that callers consumers can tell the difference between a network error or an api error.

rpcserver/router.go
jackstar12: its a bit hard since it relies on the request and some other variables defined above (like `wallet` etc.)
pulling the routing hint stuff into its own function wouldnt be an issue tough

michael1011: Then let's start with that. I don't like that this function in the router keeps growing

michael1011 (approved):

Issue #122, opened 2024-03-20
Add new independent endpoints for tx broadcasting and reading the chain


kilrau:
Let's broadcast via all of these by default to eleminate trust in Boltz for broadcasting transactions
1. [mempool.space](https://mempool.space/docs/api/rest#post-transaction)
2. [blockstream.info](https://github.com/Blockstream/esplora/blob/master/API.md#post-tx)
3. [Boltz](https://api.boltz.exchange/swagger)

Same for web app: https://github.com/BoltzExchange/boltz-web-app/issues/521

PR #147, opened 2024-05-04
refactor: simplify internal onchain interfaces

jackstar12:
The usage of wallets / ln nodes as providers for fee estimations / block updates always involved additional complexity which doesnt really pay off since we use default mempool endpoints in testnet / mainnet anyways, while power users can still configure their own mempool / electrum server.

closes: #122


cmd/boltzd/boltzd.go
```diff
@@ -217,17 +213,23 @@ func initOnchain(cfg *config.Config, network *boltz.Network) (*onchain.Onchain,
...
 		logger.Info("liquid.network API: " + cfg.MempoolLiquidApi)
-		mempoolLiquid := mempool.InitClient(cfg.MempoolLiquidApi)
-		onchain.Liquid.Fees = mempoolLiquid
-		onchain.Liquid.Listener = mempoolLiquid
+		client := mempool.InitClient(cfg.MempoolLiquidApi)
+		chain.Liquid.Blocks = client
+		chain.Liquid.Tx = client
+	}
+
+	// use boltz in regtest to avoid situations where electrum doesnt know about the tx yet
```
michael1011, line 228:
```suggestion
	// Use the Boltz API in regtest to avoid situations where electrum doesnt know about the tx yet
```

michael1011 (approved):

PR #153, opened 2024-05-07
fix: round up fee by 1

jackstar12:
Rounding up instead of down makes sure we don't go below the minrelayfee when using the min sat/vbyte.


boltz/transaction.go
```diff
@@ -117,7 +117,7 @@ func ConstructTransaction(network *Network, currency Currency, outputs []OutputD
 		return nil, nil, err
 	}
 
-	fee := uint64(float64(noFeeTransaction.VSize()) * satPerVbyte)
+	fee := uint64(float64(noFeeTransaction.VSize())*satPerVbyte) + 1
```
michael1011, line 120: What about using a [`math.Ceil`](https://pkg.go.dev/math#Ceil) instead of + 1

michael1011 (approved):

PR #154, opened 2024-05-07
refactor: do not implicitly use nil entity as admin default

jackstar12:


autoswap/autoswap_test.go
```diff
@@ -1,6 +1,8 @@
 package autoswap
 
 import (
+	"fmt"
+	rand2 "golang.org/x/exp/rand"
```
michael1011, line 5: Why the alias?

boltzrpc/boltzrpc.proto
```diff
@@ -258,6 +258,7 @@ message SwapInfo {
     optional uint64 onchain_fee = 18;
     // internal wallet which was used to pay the swap
     optional int64 wallet_id = 20;
+    int64 entity_id = 21;
```
michael1011, line 261: Those ids can't be negative, can they? Why not use `uint64`?

database/entity.go
```diff
@@ -9,6 +9,19 @@ type Entity struct {
 	Name string
 }
 
+const DefaultEntityId Id = 1
+const DefaultEntityName string = "default"
```
michael1011, line 13: Should we call it admin?

jackstar12: makes sense

michael1011 (approved):
utACK

PR #157, opened 2024-05-12
feat: chain autoswap budget

jackstar12:


database/stats.go
```diff
@@ -6,16 +6,20 @@ import (
...
       FROM swaps
       UNION ALL
-      SELECT serviceFee, onchainFee + (routingFeeMsat / 1000), expectedAmount, isAuto, createdAt, false isChain
+      SELECT serviceFee, onchainFee + (routingFeeMsat / 1000), expectedAmount, isAuto, createdAt, false isChain, entityId
       FROM reverseSwaps
-      ) stats
+      UNION ALL
+      SELECT serviceFee, onchainFee, data.amount, isAuto, createdAt, true isChain, entityId
+      FROM chainSwaps
+      JOIN chainSwapsData data ON chainSwaps.id =data.id AND data.currency = chainSwaps.fromCurrency
```
michael1011, line 17:
```suggestion
      JOIN chainSwapsData data ON chainSwaps.id = data.id AND data.currency = chainSwaps.fromCurrency
```

database/stats.go
```diff
@@ -6,16 +6,20 @@ import (
...
+      ) stats 
 `
 
 func (database *Database) QueryStats(args SwapQuery, chainOnly bool) (*boltzrpc.SwapStats, error) {
```
michael1011, line 21: `chainOnly` implies that setting it to false means that all swap types are included. What it actually does is either chain or submarine and reverse

jackstar12: what about a `swapTypes []boltz.SwapType` as an argument?

autoswap/autoswap_test.go
```diff
@@ -13,7 +13,9 @@ import (
 
 func getTestDb(t *testing.T) *database.Database {
 	db := &database.Database{
-		Path: ":memory:",
+		// its important to use a shared cache here since the sql driver can open up multiple connections
```
michael1011, line 16:
```suggestion
		// it's important to use a shared cache here since the SQL driver can open up multiple connections
```
That' an important one :partying_face:

database/budget.go
```diff
@@ -2,32 +2,42 @@ package database
...
 type BudgetInterval struct {
 	StartDate time.Time
 	EndDate   time.Time
+	Name      string
+	EntityId  Id
 }
 
-func (database *Database) QueryCurrentBudgetInterval() (*BudgetInterval, error) {
-	row := database.QueryRow("SELECT * FROM autobudget ORDER BY startDate DESC LIMIT 1")
+func (database *Database) QueryCurrentBudgetInterval(name string, entityId Id) (*BudgetInterval, error) {
```
michael1011, line 16: Why not use the `swapperType` as parameter of the function for type safety and cast to string in the query?

database/stats.go
michael1011: Still not fixed

michael1011 (approved):
utACK

database/stats.go
michael1011: That'd be best imho

database/budget.go
jackstar12: There are cyclic dependency issues and I'm not sure about moving the `SwapperType` into the db package

PR #167, opened 2024-06-26
feat: allow manual refunds to wallets

jackstar12:
- feat: manual refunds to wallets
- refactor: use pointers instead of values in db functions


michael1011 (approved):
utACK

PR #171, opened 2024-06-27
feat: make sure all gdk accounts are synced on startup

jackstar12:


onchain/wallet/handler.go
```diff
@@ -11,12 +11,12 @@ import "C"
...
+	subaccountNotification Notification = "subaccount"
 )
 
 type handlerFunc = func(map[string]any)
 
-var handlers = make(map[string]handlerFunc)
+var handlers = make(map[Notification]handlerFunc)
 var handlerLock = sync.Mutex{}
 
 // this has to be in a seperate file to avoid c redefition issues
```
michael1011, line 22:
```suggestion
// this has to be in a separate file to avoid c redefinition issues
```

onchain/wallet/wallet.go
```diff
@@ -109,6 +112,14 @@ type Config struct {
 
 var config *Config
 
+type syncListener struct {
+	accounts []uint64
+	done     chan bool
+}
+
+var syncListeners []*syncListener
+var syncListenersLock = sync.Mutex{}
```
michael1011, line 121: These should be declared as fields of `Wallet`

onchain/wallet/wallet.go
```diff
@@ -308,33 +345,23 @@ func (wallet *Wallet) GetSubaccounts(refresh bool) ([]*Subaccount, error) {
...
-			}
+	var newAccounts []uint64
+	for _, account := range result.Subaccounts {
+		if !slices.Contains(wallet.syncedAccounts, account.Pointer) {
+			newAccounts = append(newAccounts, account.Pointer)
+			wallet.syncedAccounts = append(wallet.syncedAccounts, account.Pointer)
 		}
 	}
+	if len(newAccounts) > 0 {
+		done := make(chan bool)
```
michael1011, line 356: This channel is never closed

rpcserver/router.go
```diff
@@ -1382,29 +1384,44 @@ func (server *routedBoltzServer) unlock(password string) error {
...
 	}
-	for _, creds := range credentials {
-		wallet, err := wallet.Login(creds)
+
+	server.syncing = true
+	go func() {
+		defer func() {
+			server.syncing = false
+		}()
+		wg := sync.WaitGroup{}
```
michael1011, line 1393:
```suggestion
		var wg sync.WaitGroup
```

rpcserver/router.go
```diff
@@ -1382,29 +1384,44 @@ func (server *routedBoltzServer) unlock(password string) error {
...
+		for _, creds := range credentials {
+			creds := creds
+			wg.Add(1)
```
michael1011, line 1396:
```suggestion
                wg.Add(len(credentials))
		for _, creds := range credentials {
			creds := creds
```

onchain/wallet/wallet.go
jackstar12: No, they have to be global, since the gdk notification unfortunately is global aswell and doesn't contain information about the wallet.

onchain/wallet/wallet.go
jackstar12: is this a problem? gc picks it up even if it doesnt get closed

michael1011 (approved):

PR #178, opened 2024-07-08
chore: cleanup cli

jackstar12:


kilrau (approved):
- `boltzcli listswaps` is good
- `createswap` & `createreverseswap` good too
- fee % rounding error fixed too

PR #181, opened 2024-07-08
fix: How do you want to specify min/max balance values? -> How do you…

kilrau:
… want to specify values?


cmd/boltzcli/commands.go
```diff
@@ -722,7 +722,7 @@ func autoSwapLightningSetup(ctx *cli.Context) error {
 
 	var balanceType string
 	prompt = &survey.Select{
-		Message: "How do you want to specify min/max balance values?",
+		Message: "How do you want to specify values?",
```
jackstar12, line 725:
```suggestion
		Message: "How do you want to specify balance values?",
```

PR #182, opened 2024-07-08
fix: ignore other threshold if only one swap type is allowed

jackstar12:


autoswap/lightning_test.go
```diff
@@ -428,7 +428,7 @@ func TestStrategies(t *testing.T) {
 			config: &SerializedLnConfig{
 				PerChannel:             true,
 				InboundBalancePercent:  40,
```
kilrau, line 430:
```suggestion
				InboundBalancePercent:  40, // inbound is ignored if only normal swaps are executed
```

kilrau (approved):

PR #192, opened 2024-07-10
ci: sign releases and add verification instructions

jackstar12:


.github/workflows/release.yml
```diff
@@ -36,5 +36,33 @@ jobs:
...
+            bin/*.tar.gz
+            boltz-client-manifest-*.txt
           draft: true
+          body: |
+            # Verifying the Release
+
+            In order to verify the release, you'll need to have `gpg` or `gpg2` installed on your system. You'll first need to import the keys that have signed this release if you haven't done so already: 
+
+            ```
+            curl https://canary.boltz.exchange/pgp.asc | gpg --import
```
michael1011, line 49: Let's put that key into this repo too and download via raw github links

.github/workflows/release.yml
```diff
@@ -36,5 +36,33 @@ jobs:
...
+            You should see the following if the verification was successful:
+
+            ```
+            gpg: Signature made Mo 01 Jul 2024 00:40:51 CEST
+            gpg:                using RSA key C2640F630570F5EDEDE02DE684D249BA71685D46
+            gpg: Good signature from "Michael <me@michael1011.at>" [unknown]
+            gpg:                 aka "Michael <michael101101@me.com>" [unknown]
+            ```
+
+            This will verify the signature of the boltz-client-manifest file, which ensures integrity and authenticity of the archive you've downloaded locally containing the binaries. Next, depending on your operating system, you should then re-compute the `sha256` hash of the archive with `shasum -a 256 <filename>`, compare it with the corresponding one in the boltz-client-manifest file, and ensure they match *exactly*.
```
michael1011, line 67: I always use `sha256sum -c`. What's the difference?

jackstar12: that i did not know about `-c`

jackstar12: The `-c` option expects all files to be available in the same order as in the manifest which won't be the case if you just download one archive

michael1011 (approved):

kilrau (approved):

PR #198, opened 2024-07-10
feat: claimswaps rpc

jackstar12:
The `ClaimSwaps` rpc (`claimswaps` in cli) allows claiming multiple swaps to an address or internal wallet.
Its required when a reverse or chain swap has not address or wallet set. Right now, this can only happen if the wallet of a swap is removed after its creation.


boltzrpc/boltzrpc.proto
```diff
@@ -39,6 +39,12 @@ service Boltz {
     */
     rpc RefundSwap (RefundSwapRequest) returns (GetSwapInfoResponse);
 
+    /*
+    Claim a swap manually.
+    This is only required when no claim address has been set or the swap does not have an associated wallet.
```
michael1011, line 44:
```suggestion
    This is only required when no claim address has been set and the swap does not have an associated wallet.
```
Both have to be true, no?

nursery/refund.go
```diff
@@ -64,3 +64,14 @@ func (nursery *Nursery) RefundSwaps(currency boltz.Currency, swaps []*database.S
 	}
 	return nursery.createTransaction(currency, outputs)
 }
+
+func (nursery *Nursery) ClaimSwaps(currency boltz.Currency, reverseSwaps []*database.ReverseSwap, chainSwaps []*database.ChainSwap) error {
```
michael1011, line 68: Why is this in a file called `refund.go`?

boltzrpc/boltzrpc.proto
```diff
@@ -407,6 +416,14 @@ message RefundSwapRequest {
     }
 }
 
+message ClaimSwapRequest {
```
michael1011, line 419: What about allowing multiple swaps to be claimed to the same address or wallet in one transaction?

jackstar12: Can be done

michael1011: > What should the response be in that case? All swapinfos or just the claim transaction id

ID and claim transaction id should be enough.

> Its a inconvenient since it should probably align with refundswap which is already included in the last release

Yes, but let's not constrain ourselves by past mistakes. I'd rather have them be different than both not optimal

cmd/boltzd/boltzd_test.go
```diff
@@ -227,8 +228,10 @@ func setup(t *testing.T, options setupOptions) (client.Boltz, client.AutoSwap, f
 		require.NoError(t, err)
 		_, err = autoSwapClient.ResetConfig(client.ChainAutoSwap)
 		require.NoError(t, err)
-		_, err = boltzClient.RemoveWallet(testWallet.Id)
-		require.NoError(t, err)
+		/*
+			_, err = boltzClient.RemoveWallet(testWallet.Id)
+			require.NoError(t, err)
+		*/
```
michael1011, line 234: rm

cmd/boltzd/test/wallet/registry/elements-regtest/icons
```diff
@@ -0,0 +1 @@
+�
```
michael1011, line 1: What about all these new empty files being committed?

michael1011 (changes requested):

cmd/boltzd/test/wallet/registry/elements-regtest/icons
jackstar12: accident

Issue #189, opened 2024-07-09
Autoswap: Add mainchain miner fee level as trigger for chain swaps


kilrau:

jackstar12:
We basically have this through `maxFeePercent`

kilrau:
Yes but need to expose it in `setup`. And in an understandable manner

PR #199, opened 2024-07-10
feat: add more validators and `maxFeePercent` to autoswap setup

jackstar12:
closes: #189


cmd/boltzcli/utils.go
```diff
@@ -114,3 +114,22 @@ func printDeposit(amount uint64, address string, hours float32, blockHeight uint
 	fmt.Printf("Please deposit %s into %s in the next ~%.1f hours (block height %d)\n",
 		amountString, address, hours, blockHeight)
 }
+
+func uintValidator(ans interface{}) error {
+	_, err := strconv.ParseUint(ans.(string), 10, 64)
```
michael1011, line 119: Are those casts safe to do?

jackstar12: As long as you dont try to use them in a multi select (answer is a []string in that case) - yes

michael1011 (approved):

PR #207, opened 2024-07-11
feat: remember time of reverse swap payment

jackstar12:


boltzrpc/boltzrpc.proto
```diff
@@ -319,6 +319,7 @@ message ReverseSwapInfo {
     repeated ChannelId chan_ids = 15;
     optional string blinding_key = 16;
     int64 created_at = 17;
+    int64 paid_at = 23;
```
michael1011, line 322:
```suggestion
    optional int64 paid_at = 23;
```

This is null when the invoice has not been paid yet

database/database.go
```diff
@@ -126,6 +126,7 @@ CREATE TABLE reverseSwaps
     serviceFeePercent   REAL    DEFAULT 0,
     onchainFee          INT,
     createdAt           INT,
+    paidAt 			    INT,
```
michael1011, line 129: Get rid of the tab in there

database/reverse.go
```diff
@@ -200,6 +202,7 @@ func parseReverseSwap(rows *sql.Rows) (*ReverseSwap, error) {
 	}
 
 	reverseSwap.CreatedAt = parseTime(createdAt.Int64)
+	reverseSwap.PaidAt = parseTime(paidAt.Int64)
```
michael1011, line 205: Should we really coalesce this with 0?

jackstar12: Not that critical since its just internally

michael1011 (approved):

PR #215, opened 2024-07-19
fix: automatically try to resync wallet subaccounts when not found

jackstar12:


onchain/wallet/wallet.go
```diff
@@ -470,7 +470,20 @@ func Login(credentials *Credentials) (*Wallet, error) {
 
 	if credentials.Subaccount != nil {
 		if _, err := wallet.SetSubaccount(credentials.Subaccount); err != nil {
-			logger.Errorf("Failed to set subaccount for wallet %s. You might have to resync it using GetSubaccounts: %v", credentials.Name, err)
+			logger.Infof("Failed to set subaccount for wallet %s, Resyncing subaccounts", credentials.Name)
```
michael1011, line 473: How about a warning instead?

michael1011 (approved):
utACK

Issue #168, opened 2024-06-27
rpc: respond with a meaningful error message 


kilrau:
Would be nice to have the rpc server responding with a nicer error message like "Connect/unlock your Lightning node first" here:

```
Jun 27 11:50:58 K-Yoga boltzd[26233]: 2024-06-27 11:50:58.715 [ERROR] Could not get lightning info: rpc error: code = Unknown desc = wallet locked, unlock it to enable full RPC access
Jun 27 11:50:58 K-Yoga boltzd[26233]: 2024-06-27 11:50:58.715 [INFO ] Retrying in 15 seconds
kilrau@K-Yoga:~/.boltz$ boltzcli getinfo
could not connect to boltzd. make sure it is running at 127.0.0.1:9002 and try again
```

PR #216, opened 2024-07-19
feat: improve rpc startup error states

jackstar12:
- refactor: move startup responsibility to server to allow for earlier startup and better rpc errors
- test: mock syncing lightning node

closes: #168


michael1011 (approved):
utACK

Issue #214, opened 2024-07-18
rpc: GetStats


jackstar12:
Allow retrieving global swap stats, just like autoswap

PR #222, opened 2024-07-26
feat: stats rpc

kilrau:
Closes https://github.com/BoltzExchange/boltz-client/issues/214. It's a dependency for https://github.com/BoltzExchange/btcpayserver-boltz, so I'd love to get this merged soon.


boltzrpc/boltzrpc.proto
```diff
@@ -409,6 +415,15 @@ message ListSwapsResponse {
     repeated ChainSwapInfo chain_swaps = 4;
 }
 
+message GetStatsRequest {
+    // only consider manual or autoswaps. returns both if left empty
+    optional bool is_auto = 1;
```
michael1011, line 420: I'd prefer an enum. Optional values in gRPC are weird in some languages

boltzrpc/boltzrpc.proto
```diff
@@ -33,6 +33,11 @@ service Boltz {
     */
     rpc ListSwaps (ListSwapsRequest) returns (ListSwapsResponse);
 
+    /*
+    Returns a list of all swaps, reverse swaps, and chain swaps in the database.
```
michael1011, line 37: This is not what the call does

boltzrpc/boltzrpc.proto
```diff
@@ -592,9 +607,10 @@ message LightningChannel {
 message SwapStats {
     uint64 total_fees = 1;
     uint64 total_amount = 2;
-    uint64 avg_fees = 3;
-    uint64 avg_amount = 4;
+    uint64 avg_fees = 3 [deprecated = true];
+    uint64 avg_amount = 4 [deprecated = true];
```
michael1011, line 611: Why are those deprecated?

database/stats.go
```diff
@@ -34,13 +58,15 @@ func (database *Database) QueryStats(args SwapQuery, swapTypes []boltz.SwapType)
...
-	err := rows.Scan(&stats.TotalFees, &stats.TotalAmount, &stats.Count)
+	err := rows.Scan(&stats.TotalFees, &stats.TotalAmount, &stats.Count, &stats.SuccessCount)
 	if err != nil {
 		return nil, err
 	}
-	if stats.Count != 0 {
-		stats.AvgFees = stats.TotalFees / stats.Count
-		stats.AvgAmount = stats.TotalAmount / stats.Count
+	if stats.SuccessCount != 0 {
+		//nolint:staticcheck
```
michael1011, line 66: Why?

database/stats.go
```diff
@@ -8,17 +8,41 @@ import (
...
+       COALESCE(SUM(CASE WHEN state == 1 THEN amount END), 0),
+       COUNT(*),
+       COUNT(CASE WHEN state == 1 THEN 1 END)
+FROM (SELECT state,
+             serviceFee,
+             onchainFee,
+             expectedAmount amount,
+             isAuto,
+             createdAt,
+             'submarine'    type,
```
michael1011, line 21: The indentation of this statement is weird

boltzrpc/boltzrpc.proto
```diff
@@ -592,9 +607,10 @@ message LightningChannel {
 message SwapStats {
     uint64 total_fees = 1;
```
michael1011, line 608: I'd love for this to be much more granular in general. Just returning sums of stuff is not particularly useful when you consider it's 3 different swap types on 2 different chains. What actionable insight does that sum even give you in that context?

michael1011 (changes requested):

boltzrpc/boltzrpc.proto
jackstar12: > I'd love for this to be much more granular in general.

We could either return it grouped by swap type or allow for more filtering in the request, like we do with `ListSwaps`

> What actionable insight does that sum even give you in that context?

Just a general overview. For example, btcpay plugin users who use the client to receive lightning payments are probably interested in how much fees they are paying compared to the total volume.

database/stats.go
jackstar12: its goland formatting, not sure

boltzrpc/boltzrpc.proto
jackstar12: initially did it because they are just derived from the other ones and a bit of unnecessary. but we could still keep them

boltzrpc/boltzrpc.proto
jackstar12: I think its fine, we have the same in `ListSwaps` and optional values in general all over the place. How would you name the enum?

michael1011: - `ALL`
- `AutoSwap`
- `Manual`

boltzrpc/boltzrpc.proto
```diff
@@ -409,6 +419,15 @@ message ListSwapsResponse {
     repeated ChainSwapInfo chain_swaps = 4;
 }
 
+message GetStatsRequest {
+    // which swaps to include in the stats
```
michael1011, line 423: I think this doesn't need the comment

boltzrpc/boltzrpc.proto
```diff
@@ -409,6 +419,15 @@ message ListSwapsResponse {
...
+    IncludeSwaps include = 1;
```
michael1011, line 424: How about making it optional and defaulting to all?

database/stats.go
michael1011: Still, look at it in github. Super weird

database/database.go
```diff
@@ -285,9 +285,9 @@ func (query *SwapQuery) ToWhereClauseWithExisting(conditions []string, values []
 		}
 		conditions = append(conditions, "state IN ("+strings.Join(states, ",")+")")
 	}
-	if query.IsAuto != nil {
+	if query.Include != boltzrpc.IncludeSwaps_ALL {
```
michael1011, line 288: Where is the clause for manual only?

boltzrpc/boltzrpc.proto
jackstar12: The default value is ALL anyways, even if not optional, so I would leave it like this. Optional values are a bit annoying to deal with and I think its fine in this case.

database/database.go
jackstar12: if its not all then we query by `isAuto == query.Include == boltzrpc.IncludeSwaps_AUTO` - so we cover both in that branch

michael1011 (approved):

PR #224, opened 2024-07-29
chore: update regtest

jackstar12:


.gitmodules
```diff
@@ -2,3 +2,7 @@
 	path = go-secp256k1-zkp
 	url = https://github.com/vulpemventures/go-secp256k1-zkp
 	branch = v1.1.6
+[submodule "regtest"]
+	path = regtest
+	url = https://github.com/BoltzExchange/regtest
+	branch = main
```
michael1011, line 8: Why a submodule and not `git clone` or `git pull` when needed? Feels like this is a risk of it becoming outdated easily

rpcserver/test/boltz.toml
```diff
@@ -1,70 +1,14 @@
...
-
-# Whether the macaroon authentication for the gRPC and REST interface should be disabled
-noMacaroons = false
-
-# Path to the admin macaroon for the gRPC and REST interface
-adminMacaroonPath = ""
-
-# Path to the read only macaroon for the gRPC and REST interface
-readOnlyMacaroonPath = ""
+datadir = "../regtest/data/cln1"
```
michael1011, line 14: Missing EOF

PR #234, opened 2024-08-01
feat: use lowball fee on liquid

jackstar12:


onchain/wallet/wallet.go
```diff
@@ -295,7 +294,9 @@ func (wallet *Wallet) Connect() error {
 		return err
 	}
 
-	params := make(map[string]any)
+	params := map[string]any{
+		"min_fee_rate": MinFeeRate * 1000,
```
michael1011, line 298: Maybe add a comment that GDK uses sat/kVB

michael1011 (approved):

PR #241, opened 2024-08-12
feat: lowball flags for lockup transactions of submarine and chain swaps

jackstar12:


boltzrpc/boltzrpc.proto
```diff
@@ -484,6 +484,9 @@ message CreateSwapRequest {
     optional uint64 wallet_id = 6;
     // invoice to use for the swap. if not set, the daemon will get a new invoice from the lightning node
     optional string invoice = 7;
+    // make sure boltz will accept the lockup transaction without waiting for a confirmation if the amount is within limits,
+    // involves higher fees for liquid
```
michael1011, line 488:
```suggestion
    // Boltz does not accept 0-conf for Liquid transactions with a fee of 0.01 sat/vByte;
    // when `zero_conf` is enabled, a fee of 0.1 sat/vByte will be used for Liquid lockup transactions
```

michael1011 (approved):

Issue #229, opened 2024-07-31
error: Could not set swap service fee in database


kilrau:
![image](https://github.com/user-attachments/assets/0a96f5eb-0926-4ab9-a26c-30a74075cd08)

PR #242, opened 2024-08-12
fix: correct service fee calculation for submarine swaps

jackstar12:
closes: #229

### DIFF
--- a/nursery/swap.go
+++ b/nursery/swap.go
@@ -288,12 +288,16 @@ func (nursery *Nursery) handleSwapStatus(swap *database.Swap, status boltz.SwapS
 			return
 		}
 		invoiceAmount := uint64(decodedInvoice.MilliSat.ToSatoshis())
-		serviceFee := swap.ServiceFeePercent.Calculate(swap.ExpectedAmount)
-		boltzOnchainFee := swap.ExpectedAmount - invoiceAmount - serviceFee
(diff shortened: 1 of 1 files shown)


nursery/swap.go
```diff
@@ -288,12 +288,16 @@ func (nursery *Nursery) handleSwapStatus(swap *database.Swap, status boltz.SwapS
 			return
 		}
 		invoiceAmount := uint64(decodedInvoice.MilliSat.ToSatoshis())
-		serviceFee := swap.ServiceFeePercent.Calculate(swap.ExpectedAmount)
-		boltzOnchainFee := swap.ExpectedAmount - invoiceAmount - serviceFee
+		serviceFee := swap.ServiceFeePercent.Calculate(invoiceAmount)
+		boltzOnchainFee := int64(swap.ExpectedAmount - invoiceAmount - serviceFee)
+		if boltzOnchainFee < 0 {
+			logger.Warnf("Boltz onchain fee seems to be negative")
```
michael1011, line 294: How could that even happen? It shouldn't be possible at all and doing that check after casting the `uint64` result of the calculation to `int64` wouldn't even work I guess?

michael1011 (changes requested):

nursery/swap.go
michael1011: Nevermind; I am wrong

michael1011 (approved):

PR #252, opened 2024-08-19
feat: regular pings for boltz ws

jackstar12:


boltz/ws.go
```diff
@@ -85,6 +80,31 @@ func (boltz *Websocket) Connect() error {
...
+		logger.Silly("Received pong")
+		return setDeadline()
+	})
+
+	go func() {
+		ticker := time.NewTicker(pingInterval)
+		defer ticker.Stop()
+		for range ticker.C {
+			// wont wait longer for writing then for the response
+			err := conn.WriteControl(websocket.PingMessage, []byte(pingMessage), time.Now().Add(pongWait))
```
michael1011, line 97: Empty bytes in the frame do not work?

boltz/ws.go
```diff
@@ -85,6 +80,31 @@ func (boltz *Websocket) Connect() error {
...
+	conn.SetPongHandler(func(string) error {
+		logger.Silly("Received pong")
+		return setDeadline()
+	})
+
+	go func() {
+		ticker := time.NewTicker(pingInterval)
+		defer ticker.Stop()
+		for range ticker.C {
+			// wont wait longer for writing then for the response
```
michael1011, line 96:
```suggestion
			// Will not wait longer with writing than for the response
```

boltz/ws.go
```diff
@@ -85,6 +80,31 @@ func (boltz *Websocket) Connect() error {
...
+			err := conn.WriteControl(websocket.PingMessage, []byte(pingMessage), time.Now().Add(pongWait))
+			if err != nil {
+				if boltz.closed {
+					return
+				}
+				logger.Error("could not send ping, closing conn: " + err.Error())
```
michael1011, line 102:
```suggestion
				logger.Error("could not send ping; closing connection: %s", err)
```

boltz/ws.go
jackstar12: It does not

PR #258, opened 2024-08-19
feat: resubscribe to swaps when reconnecting websocket

jackstar12:


boltz/ws.go
```diff
@@ -162,6 +164,13 @@ func (boltz *Websocket) Connect() error {
 		}
 	}()
 
+	if len(boltz.swapIds) > 0 {
+		// Clear swapIds to avoid duplicates but still resubscribe to the same swaps.
```
michael1011, line 168: Wouldn't using a `map[string]bool` instead of a slice solve that problem in a nicer way?

jackstar12: or just a private version of Subscribe which doesnt append to the swapIds

michael1011 (approved):

PR #264, opened 2024-08-21
feat: properly implement magic routing hints

jackstar12:


nursery/reverse.go
```diff
@@ -222,3 +225,126 @@ func (nursery *Nursery) handleReverseSwapStatus(reverseSwap *database.ReverseSwa
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	return slices.DeleteFunc(reverseSwaps, func(swap *database.ReverseSwap) bool {
+		return !swap.ExternalPay
+	}), nil
+}
+
+func (nursery *Nursery) searchClaimTx(notification wallet.TransactionNotification) error {
```
jackstar12, line 242: not sure how to name this

cmd/boltzcli/commands.go
```diff
@@ -1159,13 +1159,20 @@ func createSwap(ctx *cli.Context) error {
...
+		printDeposit(swap.ExpectedAmount, swap.Address, swap.TimeoutHours, uint64(swap.TimeoutBlockHeight), pairInfo.Limits)
 	}
 
-	fmt.Println("Swap ID:", swap.Id)
-
-	return swapInfoStream(ctx, swap.Id, false)
+	if swap.Id == "" {
+		if !externalPay {
+			fmt.Println()
+			fmt.Println("Invoice has been paid directly via onchain:", swap.TxId)
```
jackstar12, line 1168: how should we word this?

nursery/reverse.go
michael1011: I don't like the claim transaction in the name because it is for what you called "direct payments"

cmd/boltzcli/commands.go
michael1011: How about "Invoice has been paid directly to the wallet: <txId>"

wweseeooo5:
اين مال

PR #270, opened 2024-08-25
fix: use zero conf when invoice expiry is less than block time

jackstar12:


michael1011 (approved):

rpcserver/router.go
```diff
@@ -677,6 +677,10 @@ func (server *routedBoltzServer) createSwap(ctx context.Context, isAuto bool, re
 		if err != nil {
 			return nil, err
 		}
+		if decoded.Expiry().Minutes() < boltz.GetBlockTime(pair.From) {
```
michael1011, line 680: Does `Expiry()` return the expiry value in the invoice or *time until* expiry? And how about some buffer?

jackstar12: The time until expiry:
```
	// expiry specifies the timespan this invoice will be valid.
	// Optional. If not set, a default expiry of 60 min will be implied.
```
Lets go for a 10s buffer

michael1011 (approved):

PR #280, opened 2024-09-01
feat: listswaps pagination

jackstar12:
Adds a `unitfy` option to the `ListSwaps` rpc which allows using index and offset based pagination.  This is achieved using a new `allSwaps` view.

The `expectedAmount` column of the reverse swaps table was renamed to `onchainAmount` for consistency with the go code aswell.


boltzrpc/boltzrpc.proto
```diff
@@ -418,17 +419,44 @@ enum IncludeSwaps {
...
 
+message AnySwapInfo {
+    string id = 1;
+    SwapType type = 2;
+    Pair pair = 3;
+    SwapState state = 4;
+    optional string error = 5;
+    string status = 6;
+    // the "from" amount of the swap.
+    // its the expected amount sent to be to the lockup address for submarine and chain swaps and
```
michael1011, line 430:
```suggestion
    // It is the expected amount sent to be to the lockup address for submarine and chain swaps and
```

boltzrpc/boltzrpc.proto
```diff
@@ -418,17 +419,44 @@ enum IncludeSwaps {
...
+    // the invoice amount for reverse swaps.
+    // the "to" amount can be calculated by subtracting the service and network fee.
```
michael1011, line 432: Not sure if we should just make the life of the API consumer easier by doing that calculation for them. Would feel a lot more comfortable using that amount directly

boltzrpc/boltzrpc.proto
```diff
@@ -418,17 +419,44 @@ enum IncludeSwaps {
...
+
 message ListSwapsRequest {
     optional Currency from = 1;
     optional Currency to = 2;
     optional SwapState state = 4;
     IncludeSwaps include = 5;
+    optional uint64 limit = 6;
+    optional uint64 offset = 7;
+    // wether to return swaps in the shared `all_swaps` list or in the detailed lists.
+    // the `limit` and `offset` are only considered when `unify` is true.
```
michael1011, line 450: Wouldn't throwing an error be nicer when trying to paginate a non-unified call? Silently ignoring sounds like a bad idea

boltzrpc/boltzrpc.proto
```diff
@@ -418,17 +419,44 @@ enum IncludeSwaps {
...
+    optional bool unify = 8;
 }
 message ListSwapsResponse {
     repeated SwapInfo swaps = 1;
-    repeated CombinedChannelSwapInfo channel_creations = 2;
+    repeated CombinedChannelSwapInfo channel_creations = 2 [deprecated = true];
     repeated ReverseSwapInfo reverse_swaps = 3;
     repeated ChainSwapInfo chain_swaps = 4;
+    // populated when `unify` is set to true in the request
+    repeated AnySwapInfo all_swaps = 5;
```
michael1011, line 459: Why not a `oneof`? Either all the detailed infos or the `AnySwapInfo`?

cmd/boltzcli/commands.go
```diff
@@ -193,57 +196,75 @@ func listSwaps(ctx *cli.Context) error {
 		headerFmt := color.New(color.FgGreen, color.Underline).SprintfFunc()
 		columnFmt := color.New(color.FgYellow).SprintfFunc()
 
-		if len(list.Swaps) == 0 && len(list.ReverseSwaps) == 0 && len(list.ChainSwaps) == 0 {
-			fmt.Println("No swaps found")
-			return nil
-		}
+		if unify {
```
michael1011, line 199: This is always going to be true. Why keep the if?

database/database.go
```diff
@@ -20,6 +20,58 @@ import (
 
 // TODO: prepare insert statements only once
 
+const createViews =
+// language=sql
+`
+CREATE VIEW allSwaps AS
```
michael1011, line 26: Is there `CREATE OR REPLACE VIEW` in SQLite?

database/database.go
```diff
@@ -316,6 +370,15 @@ func (query *SwapQuery) ToWhereClauseWithExisting(conditions []string, values []
 	if len(conditions) > 0 {
 		where = " WHERE " + strings.Join(conditions, " AND ")
 	}
+	where += " ORDER BY createdAt DESC"
```
michael1011, line 373: Wouldn't this cause issues querying tables that have no column `createdAt`?

boltzrpc/boltzrpc.proto
jackstar12: `oneof` would be a breaking change in this case.

cmd/boltzcli/commands.go
jackstar12: Initially thought that i would make a flag for it. but i dont think thats necessary

database/database.go
jackstar12: Yes, but its only intenteded to be used with the swap related tables right now anyways, so I think its fine for now - can be changed if needed

database/database.go
jackstar12: Unfortunately not, but it seems like you can hack around it with `DROP IF EXISTS

michael1011 (approved):

PR #283, opened 2024-09-09
fix: track unspent tx outputs in gdk

jackstar12:


lnd/lnd.go
```diff
@@ -384,11 +384,12 @@ func (lnd *LND) GetBalance() (*onchain.Balance, error) {
 
 }
 
-func (lnd *LND) SendToAddress(address string, amount uint64, satPerVbyte float64) (string, error) {
+func (lnd *LND) SendToAddress(address string, amount uint64, satPerVbyte float64, sendAll bool) (string, error) {
 	response, err := lnd.client.SendCoins(lnd.ctx, &lnrpc.SendCoinsRequest{
 		Addr:        address,
 		Amount:      int64(amount),
 		SatPerVbyte: uint64(satPerVbyte),
+		SendAll:     sendAll,
```
michael1011, line 392: Shouldn't `sendAll` and `amount` be mutually exclusive?

onchain/wallet/wallet.go
```diff
@@ -642,24 +634,55 @@ func (wallet *Wallet) SearchOutput(txId, address string) (*onchain.Output, error
...
+	result := &outputs{}
+	if err := withAuthHandler(C.GA_get_unspent_outputs(wallet.session, params, handler), handler, result); err != nil {
+		return nil, err
+	}
+	for spent, _ := range wallet.spentOutputs {
+		found := false
+		for key, outputs := range result.Unspent {
+			for i, output := range outputs {
+				if output["txhash"] == spent {
+					result.Unspent[key] = append(outputs[:i], outputs[i+1:]...)
```
michael1011, line 661: Wouldn't it be easier to iterate over the outputs of the result and checking if they exist in `wallet.spentOutputs`? It's a map and we should take advantage of that

jackstar12: We want to check for the transactions which are in our map, but not in the gdk output set to make sure old ones are deleted, so we have to do it this way

michael1011 (approved):

PR #284, opened 2024-09-10
feat: allow specifying `all` tenant

jackstar12:


macaroons/service.go
```diff
@@ -56,6 +56,9 @@ func (service *Service) ValidateMacaroon(macBytes []byte, requiredPermissions []
 
 func (service *Service) validateTenant(ctx context.Context, raw string) (context.Context, error) {
 	if raw != "" {
+		if raw == "all" {
```
michael1011, line 59: How about making that a constant so that it can easily be compared to in the RPC server logic?

jackstar12: The rpcserver never needs to compare against that (it only checks for the entity id, which is nil in that case) l but it can be a const nevertheless

PR #306, opened 2024-10-10
feat: auto consolidate

jackstar12:


config/config.go
```diff
@@ -87,7 +87,8 @@ type Config struct {
 
 	ReferralId string
 
-	MaxZeroConfAmount *uint64 `long:"max-zeroconf-amount" description:"Maximum amount of sats to accept 0-conf"`
+	MaxZeroConfAmount        *uint64 `long:"max-zeroconf-amount" description:"Maximum amount of sats to accept 0-conf"`
+	AutoConsolidateThreshold uint64  `long:"auto-consolidate-threshold" description:"Threshold of UTXOs to trigger auto consolidation"`
```
michael1011, line 91:
```suggestion
	AutoConsolidateThreshold uint64  `long:"auto-consolidate-threshold" description:"Number of UTXOs that trigger auto consolidation"`
```

onchain/wallet/wallet.go
```diff
@@ -229,10 +233,15 @@ func Initialized() bool {
 }
 
 func Init(walletConfig Config) error {
-
 	if config != nil {
 		return errors.New("already initialized")
 	}
+	if walletConfig.AutoConsolidateThreshold == 0 {
+		walletConfig.AutoConsolidateThreshold = DefaultAutoConsolidateThreshold
+	}
```
michael1011, line 241: Let's apply the default in the config and don't auto consolidate when the configured number is 0

jackstar12: Not sure if we should allow to opt out of that since it makes the wallet unusable once the max has been reached

boltzrpc/boltzrpc.proto
```diff
@@ -765,6 +765,7 @@ message WalletTransaction {
     uint32 block_height = 6;
     // swaps which are related to this transaction
     repeated TransactionSwap swaps = 7;
+    bool is_consolidation = 8;
```
michael1011, line 768: Not a fan of having `TransactionType` and introducing boolean flags here. How about a `oneof` with `repeated TransactionSwap` and a single `TransactionType` and adding `CONSOLIDATION` to `TransactionType`?

onchain/wallet/wallet.go
```diff
@@ -843,3 +880,40 @@ func (wallet *Wallet) Ready() bool {
...
+				if err != nil {
+					return fmt.Errorf("new address: %v", err)
+				}
+				feeRate, err := wallet.estimateFee()
+				if err != nil {
+					return fmt.Errorf("could not get estimate fee: %v", err)
+				}
+				if _, err := wallet.SendToAddress(address, 0, feeRate, true); err != nil {
+					return fmt.Errorf("could not send: %v", err)
+				}
```
michael1011, line 914: Let's add a info log here. Something along the lines of `Consolidated {len(utxoSpend)} {symbol} UTXOs`

michael1011 (commented):
Untested concept ACK

I want to test this on regtest before merging though

onchain/wallet/wallet.go
```diff
@@ -119,10 +121,29 @@ type Wallet struct {
...
+	Network                  *boltz.Network
+	Debug                    bool
+	Electrum                 onchain.ElectrumConfig
+	AutoConsolidateThreshold uint64
+	MaxInputs                uint64
+}
+
+func (c *Config) Validate() error {
+	if c.MaxInputs == 0 {
+		c.MaxInputs = MaxInputs
```
michael1011, line 134: 0 should disable auto consolidation

jackstar12: This value isnt configurable and would create errors when sending normally aswell if jt was set to zero

michael1011: Ah, nvmd

michael1011 (approved):

PR #315, opened 2024-10-17
fix: handle multiple direct payments to same address correctly

jackstar12:


nursery/reverse.go
```diff
@@ -284,43 +286,65 @@ func (nursery *Nursery) checkExternalReverseSwaps(currency boltz.Currency, txId
...
+			})
+			slices.SortFunc(outputs, func(a, b *onchain.Output) int {
+				return cmp.Compare(a.Value, b.Value)
+			})
+			for _, output := range outputs {
+				var chosenSwap *database.ReverseSwap
+				for _, swap := range swaps {
+					if swap.OnchainAmount <= output.Value {
+						// try to go match as close possible to the output value
+						// its important that we sorted both swaps and outputs by amount above
```
michael1011, line 341: Let's go for an exact match. Direct payments are BIP-21s with exact amounts

database/reverse.go
```diff
@@ -244,6 +244,29 @@ func (database *Database) QueryReverseSwap(id string) (reverseSwap *ReverseSwap,
 	return reverseSwap, err
 }
 
+func (database *Database) QueryReverseSwapByClaimTransaction(txId string) (reverseSwap *ReverseSwap, err error) {
```
michael1011, line 247: Don't we have the same problem for chain swaps?

nursery/reverse.go
```diff
@@ -284,43 +290,76 @@ func (nursery *Nursery) checkExternalReverseSwaps(currency boltz.Currency, txId
...
+				if len(outputs) == 0 {
+					return true, nil
+				}
+				chosenOutput := outputs[0]
+				for _, output := range outputs {
+					if output.Value <= swap.OnchainAmount {
+						// try to go match as close possible to the output value
+						// its important that we sorted both swaps and outputs by amount above
+						chosenOutput = output
+					}
```
michael1011, line 360: Else break

jackstar12: doesnt matter since theyre sorted by amount

nursery/reverse.go
```diff
@@ -284,43 +290,74 @@ func (nursery *Nursery) checkExternalReverseSwaps(currency boltz.Currency, txId
...
 				return err
 			}
-			// ignore external adresses if we are looking for a specific id
-			if !checked && txId == "" {
-				outputs, err := nursery.onchain.GetUnspentOutputs(currency, swap.ClaimAddress)
-				if err != nil {
-					return err
-				}
+			if len(swaps) > 1 {
+				logger.Warnf("Multiple swaps for external claim address %s cant tell which one to use", claimAddress)
```
michael1011, line 311:
```suggestion
				logger.Warnf("Multiple swaps for external claim address %s cannot tell which one to use", claimAddress)
```

nursery/reverse.go
```diff
@@ -284,43 +290,74 @@ func (nursery *Nursery) checkExternalReverseSwaps(currency boltz.Currency, txId
...
+				if len(outputs) == 0 {
+					return true, nil
+				}
+				var chosenOutput *onchain.Output
+				for _, output := range outputs {
+					found, err := nursery.database.QueryReverseSwapByClaimTransaction(output.TxId)
+					if err == nil && found.Id != swap.Id {
+						continue
+					} else if err != nil && !errors.Is(err, sql.ErrNoRows) {
+						logger.Errorf("couldnt query swap by claim tx: %s", err)
```
michael1011, line 350:
```suggestion
						logger.Errorf("could not query swap by claim tx: %s", err)
```

PR #318, opened 2024-10-20
feat: use discountCT

jackstar12:
Adds support for discount ct and removes all lowball related code.
Only to be merged once discountct is widely deployed on mainnet


michael1011:
Needs changes for https://github.com/BoltzExchange/boltz-client/pull/361/files#diff-99bb9a4912420ebe5772b4e827280df5a010942ef20ef8d8b09da8961b81ad07R372

michael1011 (commented):
Shouldn't we also remove the lowball workarounds in the Liquid wallet, like setting `TransactionInputs`?

kilrau:
I am a bit worried this will brick sth for the BTCPay Plugin so maybe let's go with a manua/preview release of the plugin first

michael1011 (approved):
utACK but needs testing on testnet. I'd like to ship that when mainnet switches to discount

PR #325, opened 2024-11-03
feat: `GetSendFee` rpc

jackstar12:


rpcserver/router.go
```diff
@@ -1615,6 +1615,28 @@ func (server *routedBoltzServer) GetWallet(ctx context.Context, request *boltzrp
...
+		return nil, err
+	}
+	feeRate := request.GetSatPerVbyte()
+	if feeRate == 0 {
+		feeRate, err = server.onchain.EstimateFee(ownWallet.Currency, request.GetIsSwapAddress())
+		if err != nil {
+			return nil, err
+		}
+	}
+	if request.GetIsSwapAddress() && request.GetAddress() == "" {
```
michael1011, line 1630:
```suggestion
	if request.GetAddress() == "" {
```
Shouldn't we always coalesce?

onchain/wallet/wallet.go
```diff
@@ -754,11 +751,44 @@ func (wallet *Wallet) SendToAddress(address string, amount uint64, satPerVbyte f
...
+		}
+		Error string
+	}
+	if err := mapstructure.Decode(result, &createTx); err != nil {
+		return 0, 0, err
+	}
+	if createTx.Error != "" {
+		return 0, 0, errors.New(createTx.Error)
+	}
+	return createTx.Addressees[0].Satoshi, createTx.Fee, nil
```
michael1011, line 779: Is it guaranteed that the first address is *always* the address we want to send to and not the change?

jackstar12: This doesnt contain the actual outputs of the tx but only the addresses we told gdk to send to (gdk populates the amount in case of a sweep)

PR #340, opened 2024-11-25
fix: check for existing submarine swaps before creation

jackstar12:


rpcserver/router.go
```diff
@@ -778,6 +778,16 @@ func (server *routedBoltzServer) createSwap(ctx context.Context, isAuto bool, re
 
 	if swapResponse == nil {
 
+		if createSwap.Invoice != "" {
+			existing, err := server.database.QuerySwapByInvoice(createSwap.Invoice)
```
michael1011, line 782: It would be better to query by preimage hash

michael1011 (changes requested):

rpcserver/router.go
jackstar12: we dont have a preimage hash column in the db - think its fine this way.

PR #361, opened 2024-12-16
feat: multi tx provider

jackstar12:


onchain/multi_tx.go
```diff
@@ -0,0 +1,36 @@
...
+		if err == nil {
+			return hex, nil
+		}
+	}
+	return "", err
+}
+
+func (m MultiTxProvider) BroadcastTransaction(txHex string) (txId string, err error) {
+	for _, provider := range m.Providers {
+		txId, err = provider.BroadcastTransaction(txHex)
```
michael1011, line 20: Minor optimization: do it in parallel with an errgroup: https://pkg.go.dev/golang.org/x/sync/errgroup

michael1011 (approved):

PR #377, opened 2025-01-20
feat: get swap by payment hash

jackstar12:
todo:
- [x] db migration
- [x] tests


pkg/boltzrpc/boltzrpc.proto
```diff
@@ -500,6 +500,8 @@ message ClaimSwapsResponse {
 
 message GetSwapInfoRequest {
     string id = 1;
+    // only implemented for submarine swaps for now
```
michael1011, line 503:
```suggestion
    // Only implemented for submarine swaps
```

pkg/boltzrpc/boltzrpc.proto
```diff
@@ -500,6 +500,8 @@ message ClaimSwapsResponse {
 
 message GetSwapInfoRequest {
     string id = 1;
```
michael1011, line 502: This should be optional too?

pkg/boltzrpc/boltzrpc.proto
```diff
@@ -500,6 +500,8 @@ message ClaimSwapsResponse {
...
+    // only implemented for submarine swaps for now
+    optional bytes payment_hash = 2;
```
michael1011, line 504: How about a `oneof`?

jackstar12: breaking change, unless you want to add it  and deprecate the old `id`

internal/database/migration.go
```diff
@@ -547,6 +549,43 @@ func (database *Database) performMigration(tx *Transaction, oldVersion int) erro
...
+		for _, swap := range swaps {
+			if swap.Invoice != "" {
+				decoded, err := zpay32.Decode(swap.Invoice, boltz.MainNet.Btc)
+				if err != nil {
+					decoded, err = zpay32.Decode(swap.Invoice, boltz.TestNet.Btc)
+					if err != nil {
+						decoded, err = zpay32.Decode(swap.Invoice, boltz.Regtest.Btc)
+					}
+
+				}
```
michael1011, line 573: Don't we have information about the network at this point?

pkg/boltzrpc/boltzrpc.proto
michael1011, line 502: Can we mark this as deprecated?

pkg/boltzrpc/boltzrpc.proto
```diff
@@ -500,6 +500,8 @@ message ClaimSwapsResponse {
 
 message GetSwapInfoRequest {
     string id = 1;
+    // Only implemented for submarine swaps
+    optional bytes payment_hash = 2;
```
michael1011, line 504: And have a `oneof` `identifier` here?

internal/database/migration.go
jackstar12: We would have to add it as a param to the migration, not sure if i wanna do that

internal/database/database.go
```diff
@@ -108,6 +108,7 @@ CREATE TABLE swaps
     preimage            VARCHAR,
     redeemScript        VARCHAR,
     invoice             VARCHAR,
+    paymentHash         VARCHAR(64),
```
michael1011, line 111: The migration sets `NOT NULL`. Should we do that here too or is default behaviour of SQLite?

michael1011 (approved):

internal/database/database.go
jackstar12: Default allows null, so lets explicitly declare it here aswell.

michael1011 (approved):

PR #393, opened 2025-02-13
feat: lazy websocket connection

jackstar12:
only create actual websocket connection when there are pending swaps and
close it if there are no more pending swaps

### DIFF
--- a/pkg/boltz/ws.go
+++ b/pkg/boltz/ws.go
@@ -34,6 +34,7 @@ type Websocket struct {
 	closed            bool
 	dialer            *websocket.Dialer
 	swapIds           []string
+	swapIdsLock       sync.Mutex
 	reconnectInterval time.Duration
 }
 
@@ -223,11 +224,15 @@ func (boltz *Websocket) Subscribe(swapIds []string) error {
 			return err
 		}
 	}
+	boltz.swapIdsLock.Lock()
 	boltz.swapIds = append(boltz.swapIds, swapIds...)
+	boltz.swapIdsLock.Unlock()
 	return nil
 }
 
 func (boltz *Websocket) Unsubscribe(swapId string) {
+	boltz.swapIdsLock.Lock()
+	defer boltz.swapIdsLock.Unlock()
 	boltz.swapIds = slices.DeleteFunc(boltz.swapIds, func(id string) bool {
 		return id == swapId
 	})
--- a/pkg/boltz/ws.go
+++ b/pkg/boltz/ws.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"slices"
+	"sync"
 	"time"
 
 	"github.com/BoltzExchange/boltz-client/v2/internal/logger"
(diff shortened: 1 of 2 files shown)


pkg/boltz/ws.go
```diff
@@ -51,14 +60,19 @@ func (boltz *Api) NewWebsocket() *Websocket {
...
+		Updates:           make(chan SwapUpdate),
+		state:             disconnected,
+		reconnectInterval: reconnectInterval,
 	}
 }
 
 func (boltz *Websocket) Connect() error {
+	if boltz.state == closed {
+		return errors.New("websocket is closed")
+	}
```
michael1011, line 75: Why can't we connect again when a websocket was closed once?

pkg/boltz/ws.go
```diff
@@ -201,6 +219,11 @@ func (boltz *Websocket) subscribe(swapIds []string) error {
 }
 
 func (boltz *Websocket) Subscribe(swapIds []string) error {
+	if boltz.state == disconnected {
+		if err := boltz.Connect(); err != nil {
```
michael1011, line 223: There should be a check here for `len(swapIds) > 0` to make sure that we don't open and close the websocket instantly afterwards

michael1011 (changes requested):
It really feels like there should be locks on the pieces of code that touch the state of the websocket. The enum states that represent the state are not clear and the logic inside of the `Connect` function is getting way too complex and should be split up.

pkg/boltz/ws.go
jackstar12: Because the close call also causes the updates channel to be closed. Its not closed in this call to avoid a situation where we just got a message in the receiving goroutine and would then send on a closed channel

jackstar12:
@michael1011 the states actually aren't necessary if one refactors the existing logic a bit.

I also added some comments describing the close logic.

pkg/boltz/ws.go
```diff
@@ -219,20 +232,41 @@ func (boltz *Websocket) Unsubscribe(swapId string) {
...
 }
 
 func (boltz *Websocket) Close() error {
+	// setting this flag will cause the `Updates` channel to be closed
+	// in the receiving goroutine. this isn't done here to avoid a situation
+	// where we close the channel while the receiving routine is processing an incoming message
+	// and then tries to send on the closed channel.
+	boltz.updatesLock.Lock()
+	defer boltz.updatesLock.Unlock()
+	close(boltz.Updates)
```
michael1011, line 258: Why do we need to close that channel?

pkg/boltz/ws.go
```diff
@@ -24,15 +25,16 @@ type SwapUpdate struct {
...
-	swapIds       []string
+	Updates     chan SwapUpdate
+	updatesLock sync.Mutex
+
+	apiUrl            string
+	subscriptions     chan bool
+	conn              *websocket.Conn
+	closed            bool
+	dialer            *websocket.Dialer
+	swapIds           []string
```
michael1011, line 36: Now that I think of it, shouldn't there be a lock on `swapIds` to handle concurrent writes?

pkg/boltz/ws.go
jackstar12: So that other goroutines that are receiving on the channel can properly cleanup

pkg/boltz/ws.go
```diff
@@ -122,58 +130,56 @@ func (boltz *Websocket) Connect() error {
...
-				case "subscribe":
-					boltz.subscriptions <- true
-					continue
-				default:
-					logger.Warnf("unknown event: %s", response.Event)
+				if err := boltz.handleTextMessage(message); err != nil {
+					logger.Errorf("could not handle websocket message: %s", err)
 				}
+			default:
+				logger.Warnf("unknown message type: %v", msgType)
```
michael1011, line 137: One last thing: wouldn't this warn all the time when it gets pongs from the backend as response to the pings it sends

jackstar12: Pong messages are handled by the pong handler and arent by ReadMessage`

michael1011 (approved):

PR #407, opened 2025-03-31
feat: implement `IsTransactionConfirmed` for boltz

jackstar12:
we dont have to check for `ErrUnsupported` anymore since we always have boltz as a backup provider

### DIFF
--- a/internal/nursery/reverse.go
+++ b/internal/nursery/reverse.go
@@ -5,12 +5,13 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"slices"
+	"time"
+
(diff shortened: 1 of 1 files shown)


pkg/boltz/api.go
```diff
@@ -235,7 +235,8 @@ type GetTransactionRequest struct {
 }
 
 type GetTransactionResponse struct {
-	Hex string `json:"hex"`
+	Hex           string `json:"hex"`
+	Confirmations uint64 `json:"confirmations"`
```
michael1011, line 239: `confirmations` is not set when the transaction is not confirmed yet. Is that handled gracefully by it defaulting to 0?

jackstar12: yes

PR #454, opened 2025-05-30
ci: setup rust cache

jackstar12:
The rust rust action only caches the `./target` directory from the workspace root


rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "1.81.0"
```
michael1011, line 2: Why lock the Rust compiler version?

PR #497, opened 2025-07-06
feat: implement `transaction.direct` status update

jackstar12:
## Summary by CodeRabbit

* **New Features**
  * Added support for handling direct transactions in reverse swaps, including selection of suitable outputs and special handling for Liquid currency.
  * Introduced the ability to derive blinding keys for Liquid wallet addresses.
  * Implemented a new method to query reverse swaps by claim address.

* **Bug Fixes**
  * Prevented creation of multiple pending reverse swaps with the same claim address.

* **Refactor**
  * Separated fee estimation logic from amount checking for improved modularity.
  * Simplified and improved the logic for processing and checking reverse swaps.

* **Tests**
  * Added tests for direct output selection and updated tests to verify prevention of duplicate claim addresses in reverse swaps.
  * Re-enabled and updated Liquid-related test cases.

### DIFF
--- a/internal/nursery/reverse.go
+++ b/internal/nursery/reverse.go
@@ -6,10 +6,12 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"strings"
 	"time"
 
 	"github.com/BoltzExchange/boltz-client/v2/internal/lightning"
 	"github.com/BoltzExchange/boltz-client/v2/internal/onchain"
+	liquid_wallet "github.com/BoltzExchange/boltz-client/v2/internal/onchain/liquid-wallet"
 	"github.com/btcsuite/btcd/btcec/v2"
 
 	"github.com/BoltzExchange/boltz-client/v2/internal/database"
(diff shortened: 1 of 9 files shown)


internal/nursery/reverse.go
```diff
@@ -150,6 +152,47 @@ func (nursery *Nursery) handleReverseSwapStatus(reverseSwap *database.ReverseSwa
...
+			handleError("Could not find vout: " + err.Error())
+			return
+		}
+		_, err = nursery.onchain.BroadcastTransaction(directTransaction)
+		if err != nil {
+			if !strings.ContainsAny(err.Error(), "Already in block chain") {
+				handleError("Could not broadcast transaction: " + err.Error())
+				return
+			}
+		}
```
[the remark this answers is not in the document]
jackstar12: Electrum returns the `Already in block chain`, while esplora or mempool return the newer error message

internal/rpcserver/router.go
```diff
@@ -986,6 +986,18 @@ func (server *routedBoltzServer) createReverseSwap(ctx context.Context, isAuto b
 		if err != nil {
 			return nil, fmt.Errorf("invalid claim address %s: %w", claimAddress, err)
 		}
+
+		swaps, err := server.database.QueryReverseSwaps(database.SwapQuery{
+			States: []boltzrpc.SwapState{boltzrpc.SwapState_PENDING},
+		})
```
michael1011, line 992: https://github.com/BoltzExchange/boltz-client/pull/497/files#r2192366502

That should really be done. Let SQLite do the filtering for you

internal/nursery/reverse.go
```diff
@@ -235,36 +282,64 @@ func (nursery *Nursery) handleReverseSwapStatus(reverseSwap *database.ReverseSwa
...
+			continue
 		}
 
-		currency := swap.Pair.To
-		pairs, err := nursery.boltz.GetReversePairs()
-		if err != nil {
-			return fmt.Errorf("get reverse pairs: %s", err)
+		// we dont know the output value of liquid when the claim address is external
+		if swap.Pair.To == boltz.CurrencyLiquid && swap.WalletId == nil && output.Value == 0 {
+			logger.Infof("Skipping amount validation for liquid since we cant unblind external outputs")
```
michael1011, line 302:
```suggestion
			logger.Infof("Skipping amount validation for liquid since we cannot unblind external outputs")
```

internal/nursery/reverse.go
```diff
@@ -235,36 +282,64 @@ func (nursery *Nursery) handleReverseSwapStatus(reverseSwap *database.ReverseSwa
...
+			logger.Infof("Direct Transaction is claim of %s", found.Id)
+			continue
 		}
 
-		currency := swap.Pair.To
-		pairs, err := nursery.boltz.GetReversePairs()
-		if err != nil {
-			return fmt.Errorf("get reverse pairs: %s", err)
+		// we dont know the output value of liquid when the claim address is external
+		if swap.Pair.To == boltz.CurrencyLiquid && swap.WalletId == nil && output.Value == 0 {
```
michael1011, line 301: Should we just accept the payment in this case? Seems like an easy exploit
```suggestion
		// We don't know the output value of liquid when the claim address is external
		if swap.Pair.To == boltz.CurrencyLiquid && swap.WalletId == nil && output.Value == 0 {
```

michael1011 (approved):

PR #557, opened 2025-09-25
feat: hide mrh behind flag

jackstar12:


pkg/boltzrpc/boltzrpc.proto
```diff
@@ -621,6 +621,9 @@ message CreateReverseSwapRequest {
 
     // The routing fee limit for paying the lightning invoice in ppm (parts per million)
     optional uint64 routing_fee_limit_ppm = 13;
+
+    // add a magic routing hint to the lightning invoice if `external_pay` is true and an internal `wallet` is used.
```
michael1011, line 625: Specifying conditions in comments of parameters is not ok. The call should fail if the preconditions are not met

PR #571, opened 2025-10-16
fix: check for confirmation when querying claimable swaps

jackstar12:
we need to check for tx confirmation before returning a swap as
claimable. this can otherwise lead to swaps being claimed too early,
causing unnecessary errors

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate detection of claimable swaps by validating on-chain confirmations before marking swaps as claimable; claimable lists now come through the nursery component.
* **Tests**
  * Updated tests to verify claimable-swap visibility before/after block confirmations and to rely on status updates for final swap state.


michael1011 (approved):
Rabbit comment. Rest looks good

PR #576, opened 2025-10-20
fix: cli swapinfo stream

jackstar12:
- **fix: check for successfull state and not status in cli**
- **fix: flaky status checks in reverse and chain tests**

## Summary by CodeRabbit

* **Bug Fixes**
  * Swap output now shows claim transaction ID and fees only when a swap is fully successful, improving CLI clarity.
  * Wallet import prompts no longer offer an xpub option for BTC readonly/import flows; descriptions now reference a core descriptor and mnemonic.
  * Prevent duplicate claims by requiring a missing transaction ID before claiming.

* **Tests**
  * Swap streaming tests refactored to use explicit status streams for intermediate updates.
  * One on-chain test skips execution in CI environments.

* **Refactor**
  * Consolidated claim and claimable-swap logic into helper routines and simplified block-listener flow.

* **Style**
  * Added a nolint directive to an import-wallet request construction.

### DIFF
--- a/cmd/boltzcli/commands.go
+++ b/cmd/boltzcli/commands.go
@@ -1592,7 +1592,7 @@ var walletCommands = &cli.Command{
 			Name:        "credentials",
 			ArgsUsage:   "name",
 			Usage:       "Show the credentials of a wallet",
-			Description: "Shows the credentials of a wallet. These will be a core descriptor in case of a readonly wallet and a mnemonic otherwise.",
+			Description: "Shows the the core descriptor of the wallet and the mnemonic if it is a hot wallet.",
 			Action:      requireNArgs(1, showCredentials),
 		},
 		{


cmd/boltzcli/commands.go
```diff
@@ -1589,7 +1592,7 @@ var walletCommands = &cli.Command{
 			Name:        "credentials",
 			ArgsUsage:   "name",
 			Usage:       "Show the credentials of a wallet",
-			Description: "Shows the credentials of a wallet. These will be a xpub or core descriptor in case of a readonly wallet and a mnemonic otherwise.",
+			Description: "Shows the the core descriptor of the wallet and the mnemonic if its a hot wallet.",
```
michael1011:
```suggestion
			Description: "Shows the the core descriptor of the wallet and the mnemonic if it is a hot wallet.",
```

michael1011 (approved):

PR #592, opened 2025-11-20
refactor: increase default wallet sync interval

jackstar12:
mainchain syncs can take longer since we dont have waterfalls there

## Summary by CodeRabbit

* **Refactor**
  * Replaced single global wallet sync interval with per-currency sync intervals for currency-specific timing.

* **Bug Fixes**
  * Initialization now ensures per-currency defaults are set (including fast regtest defaults).
  * Sync loops use currency-specific intervals and fall back to default per-currency values when needed.

* **Tests**
  * Updated tests to validate per-currency sync timing and currency-aware wallet interactions.

### DIFF
--- a/internal/onchain/onchain.go
+++ b/internal/onchain/onchain.go
@@ -96,7 +96,14 @@ func (onchain *Onchain) Init() {
 	onchain.OnWalletChange = utils.ForwardChannel(make(chan []Wallet), 0, false)
 	onchain.syncCtx, onchain.syncCancel = context.WithCancel(context.Background())
 	if onchain.WalletSyncIntervals == nil {
-		onchain.WalletSyncIntervals = DefaultWalletSyncIntervals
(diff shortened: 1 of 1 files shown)


internal/onchain/onchain.go
```diff
@@ -120,7 +119,13 @@ func (onchain *Onchain) startSyncLoop(wallet Wallet) {
 		defer onchain.syncWait.Done()
 		for {
 			// avoid traffic spikes if a lot of wallets are using the same backend
-			sleep := time.Duration(float64(onchain.WalletSyncInterval) * (0.75 + rand.Float64()*0.5))
+			currency := wallet.GetWalletInfo().Currency
+			interval, ok := onchain.WalletSyncIntervals[currency]
+			if !ok {
+				logger.Warnf("No sync interval found for currency %s", currency)
```
michael1011: Should this have a default?

PR #604, opened 2025-11-24
fix: add migration for legacy btc wallets

jackstar12:
since its introduction the legacy wallet has been left false for all new
wallets, so we have to make sure we flag the old btc wallets once more

## Summary by CodeRabbit

* **Chores**
  * Updated database schema with a migration that modifies wallet data handling.

### DIFF
--- a/internal/database/migration.go
+++ b/internal/database/migration.go
@@ -21,7 +21,7 @@ type swapStatus struct {
 	status string
 }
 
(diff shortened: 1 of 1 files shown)


kilrau (commented):
Error is gone for me 👍  but needs a quick ACK from @michael1011 to make sure that db migration is sound

PR #605, opened 2025-11-24
refactor: liquid backend initialization

jackstar12:
connect to electrum and esplora servers lazily to not cause issues on
startup

## Summary by CodeRabbit

* **Refactor**
  * Improved blockchain client connection handling with optimized caching and time-bounded connection attempts to enhance wallet stability and reliability.
* **Style / UX**
  * Updated status text when registering block listeners to "Listening to new blocks for ..." for clearer user-facing feedback.


kilrau:
> These Connecting logs are debug level which is why you don't see them in your case

Ok fine, but then all "Connecting..." log entries should be on this log level

PR #649, opened 2026-03-05
fix: ignore unknown currencies

jackstar12:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of currency-pair parsing to ignore unknown/unsupported currencies and preserve valid mappings.
  * Safer handling of unknown event/status values to avoid spurious processing and reduce incorrect state transitions.

* **Tests**
  * Added tests verifying handling of known and unknown currency entries across submarine, reverse, and chain pair parsing and event parsing.


kilrau (commented):
Why not add tests for the unknown-status too?

michael1011 (approved):

PR #665, opened 2026-04-20
refactor: liquid wallet sync

jackstar12:
use one sync client per connection - shared clients dont scale

## Summary by CodeRabbit

* **Chores**
  * Improved wallet synchronization and connection handling for more reliable and performant on-chain syncs.
  * Added connection timeout behavior to avoid stalled syncs and reduce hangs.
  * Enhanced sync timing metrics and reduced noisy debug output to make logs clearer and more actionable.

### DIFF
--- a/internal/onchain/liquid-wallet/wallet.go
+++ b/internal/onchain/liquid-wallet/wallet.go
@@ -24,12 +24,13 @@ type Persister interface {
 
 type Wallet struct {
 	*lwk.Wollet
-	signer     *lwk.Signer
-	descriptor *lwk.WolletDescriptor
-	backend    *BlockchainBackend
-	info       onchain.WalletInfo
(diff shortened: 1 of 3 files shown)


internal/onchain/liquid-wallet/wallet.go
```diff
@@ -289,6 +265,102 @@ func (backend *BlockchainBackend) NewWallet(credentials *onchain.WalletCredentia
...
+}
+
+func (c *walletSyncClient) disconnect() {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	c.generation++
+	c.connectErr = nil
+
+	c.client = nil
```
michael1011: Is setting the client nil good enough? Or we do we have to close it somehow to free the resource ?

michael1011 (approved):

PR #670, opened 2026-04-20
feat: native cross compilation

jackstar12:
## Summary by CodeRabbit

* **Chores**
  * Expanded multi-platform build support enabling compilation for different CPU architectures
  * Optimized Docker build infrastructure with architecture-aware caching and cross-compilation toolchains
  * Improved build artifact handling and configuration management


tools/docker-build-static.sh
```diff
@@ -0,0 +1,32 @@
...
+
+set -eu
+
+target_os="${1:?target os required}"
+target_arch="${2:?target arch required}"
+
+case "$target_arch" in
+  amd64)
+    rust_target=x86_64-unknown-linux-gnu
+    tool_prefix=
```
michael1011: This implicitly assumes that the host is arm64

Dockerfile
```diff
@@ -35,7 +60,7 @@ COPY --from=builder /boltz-client/boltzcli /
 # Start a new, final image.
 FROM ubuntu:noble AS final
 
-RUN apt update && apt install ca-certificates -y && rm -rf /var/lib/apt/lists/*
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
```
michael1011: Why are we doing that now?

tools/docker-build-static.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
```
michael1011, line 1: `chmod +x` this file

.dockerignore
```diff
@@ -1,10 +1,14 @@
...
-*.a
-vendor
-bin
-target
-regtest
-boltzd
-boltzcli
-bdk/target/
-lwk/target/
+.git
```
michael1011: By ignoring `.git` the commit hash is gone from the version of binaries built

.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[target.x86_64-unknown-linux-gnu]
+linker = "gcc"
```
michael1011: This also assumes that the host is arm64

michael1011 (changes requested):
`make binaries` fails locally with `ERROR: failed to build: failed to solve: error writing layer blob: push access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed`. I *really* want to be able to build those binaries locally if need be

Dockerfile
jackstar12: So the final image is now copy-only and doesn't need to run any code - makes it faster.

michael1011: Marginally faster at the risk of bricking shit. The builder and runner images are based different linux distributions. We don't risk that

jackstar12: Alright. Downside being that host needs QEMU to emulate that on the final image, but guess thats worth it then.

michael1011 (approved):

PR #671, opened 2026-04-21
feat: deprecate gdk

jackstar12:
## Summary by CodeRabbit

* **New Features**
  * Automatic migration of legacy wallet credentials with user-facing warnings.
  * Startup warnings exposed in API GetInfo responses to surface important notices.

* **Deprecated / Removed**
  * Wallet subaccount RPC and related CLI commands removed.
  * GDK support removed — descriptor-based wallets are now required.

* **Documentation**
  * Build and wallet docs updated for descriptor-based wallets and migration guidance.


docs/grpc.md
```diff
@@ -1,7 +1,3 @@
----
-description: Full gRPC and REST API reference for boltzd covering all swap, wallet, and autoswap methods
----
```
michael1011, line 3: Keep that

internal/rpcserver/serializer.go
```diff
@@ -229,16 +227,6 @@ func serializeChainPair(pair boltz.Pair, chainPair *boltz.ChainPair) *boltzrpc.P
 }
 
 //nolint:staticcheck
```
michael1011:
```suggestion
```

internal/rpcserver/server.go
```diff
@@ -472,23 +471,6 @@ func initOnchain(cfg *config.Config, boltzApi *boltz.Api, network *boltz.Network
...
-			DataDir:                  cfg.DataDir,
-			Network:                  network,
-			Debug:                    false,
-			Electrum:                 electrumConfig,
-			AutoConsolidateThreshold: *cfg.AutoConsolidateThreshold,
-		})
-		if err != nil {
-			return nil, fmt.Errorf("could not init wallet: %v", err)
-		}
-	}
```
michael1011, line 490: Why do we not need that anymore? How does LWK handle it?

internal/rpcserver/router.go
```diff
@@ -2106,6 +2075,15 @@ func (server *routedBoltzServer) unlock(password string) error {
 			return err
 		}
 	}
+	migrated, warnings := server.migrateWalletCredentials(credentials)
+	server.walletMigrationWarnings = warnings
```
michael1011: These warnings are not updated after the wallets are removed

michael1011 (changes requested):
This is not deprecation. This is flat out removal. Which is fine, but we should make sure to communicate that correctly

internal/rpcserver/server.go
jackstar12: This is GDK specific configuration. The LWK backend is initialized here:
https://github.com/BoltzExchange/boltz-client/blob/d20135c5445a37ca9dc06fab5f0d9175afae7bcf/internal/rpcserver/server.go#L235-L246
Although I just noticed that it doesn't pass on the configuration override.

michael1011 (approved):

PR #672, opened 2026-04-22
refactor: wallet unique constrant

jackstar12:
only enforce same wallets within tenants

## Summary by CodeRabbit

* **Improvements**
  * Wallet credentials are now scoped per tenant, allowing the same credentials to be imported for different tenants without conflicts.
* **Bug Fixes**
  * Re-import of identical credentials is rejected only when they belong to the same tenant.
* **Chores**
  * Database migration added to apply tenant-scoped wallet uniqueness.
* **Tests**
  * Import tests updated to validate tenant-scoped duplicate-credential behavior.

### DIFF
--- a/internal/database/database.go
+++ b/internal/database/database.go
@@ -223,7 +223,7 @@ CREATE TABLE wallets
     lastIndex      INT,
 
     UNIQUE (name, tenantId, nodePubkey),
-    UNIQUE (xpub, coreDescriptor, mnemonic, nodePubkey)
+    UNIQUE (tenantId, xpub, coreDescriptor, mnemonic, nodePubkey)
 );
 CREATE TABLE tenants
 (
--- a/internal/database/migration.go
+++ b/internal/database/migration.go
@@ -643,7 +643,7 @@ func (database *Database) performMigration(tx *Transaction, oldVersion int) erro
 			mnemonic       VARCHAR,
 			subaccount     INT,
 			salt           VARCHAR,
-			tenantId       INT REFERENCES tenants (id),
+			tenantId       INT NOT NULL REFERENCES tenants (id),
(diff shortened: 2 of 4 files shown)


internal/database/migration.go
```diff
@@ -628,6 +628,36 @@ func (database *Database) performMigration(tx *Transaction, oldVersion int) erro
 		if _, err := tx.Exec(migration); err != nil {
 			return err
 		}
+	case 18:
+		logMigration(oldVersion)
+
+		migration := `
+		ALTER TABLE wallets RENAME TO old_wallets;
+		CREATE TABLE wallets
```
michael1011: I think this will blow up in our faces because of `legacy_alter_table`: https://www2.sqlite.org/draft/matrix/lang_altertable.html

michael1011 (changes requested):

internal/database/database.go
michael1011, line 221: Here `tenantId` is `NOT NULL`

internal/database/migration.go
```diff
@@ -628,6 +628,36 @@ func (database *Database) performMigration(tx *Transaction, oldVersion int) erro
...
+			id             INTEGER PRIMARY KEY AUTOINCREMENT,
+			name           VARCHAR,
+			currency       VARCHAR,
+			nodePubkey     VARCHAR,
+			xpub           VARCHAR,
+			coreDescriptor VARCHAR,
+			mnemonic       VARCHAR,
+			subaccount     INT,
+			salt           VARCHAR,
+			tenantId       INT REFERENCES tenants (id),
```
michael1011: In the migration it can be `NULL`

michael1011 (changes requested):

michael1011 (approved):

PR #684, opened 2026-05-24
fix: clamp underflow in CheckAmounts

wactario:
The two uint64 subtractions in CheckAmounts wrapped around 2^64 whenever boltz returned an ExpectedAmount smaller than invoiceAmount + service fee, producing a spurious "onchain fee way above expectation: 1844..." error. The swap was already created on the boltz server at that point but the DB row was never written, orphaning it.

Use a saturatingSub helper so a favorable quote yields networkFees=0 and checkTolerance reads it as no overage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed fee-calculation edge case that could produce incorrect (wrapped or negative) network fee values for favorable or under-quoted amounts, improving swap fee accuracy and preventing erroneous rejections.

* **Tests**
  * Added regression tests covering favorable/under-quoted scenarios to ensure fee calculations remain correct across swap types and network conditions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/BoltzExchange/boltz-client/pull/684?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/boltz/fees.go
+++ b/pkg/boltz/fees.go
@@ -73,14 +73,26 @@ func RequiredEstimations(swapType SwapType, pair Pair) []Currency {
 	return currencies
 }
 
-func CheckAmounts(swapType SwapType, pair Pair, sendAmount uint64, receiveAmount uint64, serviceFee Percentage, estimations FeeEstimations, includeClaim bool) error {
-	totalFees := sendAmount - receiveAmount
-	networkFees := totalFees
-	if swapType == NormalSwap {
-		networkFees -= CalculatePercentage(serviceFee, receiveAmount)
-	} else {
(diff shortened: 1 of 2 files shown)


wactario:
Fixes the following error:

```
BoltzQuote BTC run failed: <AioRpcError of RPC that terminated with:
	status = StatusCode.UNKNOWN
	details = "onchain fee way above expectation: 18446744073709541918 > 302+1500"
	debug_error_string = "UNKNOWN:Error received from peer  {grpc_message:"onchain fee way above expectation: 18446744073709541918 > 302+1500", grpc_status:2}"
>
```

michael1011 (approved):

PR #687, opened 2026-06-11
feat: exclude permant errors from automatic retry

jackstar12:
## Summary by CodeRabbit

* **Bug Fixes**
  * Repeat `ClaimSwaps` calls for successful manual-claims now re-broadcast successfully instead of returning “not found”.
  * Permanent broadcast errors now stop retries immediately.
  * Claimable swap lookups now exclude swaps marked with permanent errors.
  * `ClaimSwaps` now requires explicit swap IDs, returns “not found” when none match, and enforces destination-currency consistency.
* **Tests**
  * Added coverage for permanent broadcast failure stopping.
  * Added/updated tenant ownership validation scenarios.

### DIFF
--- a/internal/nursery/listener.go
+++ b/internal/nursery/listener.go
@@ -2,6 +2,8 @@ package nursery
 
 import (
 	"fmt"
+	"slices"
+	"strings"
 
 	"github.com/BoltzExchange/boltz-client/v2/internal/database"
 	"github.com/BoltzExchange/boltz-client/v2/internal/logger"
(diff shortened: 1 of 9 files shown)


internal/nursery/listener.go
```diff
@@ -19,6 +21,12 @@ func (nursery *Nursery) checkClaimableSwaps(currency boltz.Currency) error {
 	if err != nil {
 		return fmt.Errorf("could not query claimable Swaps: %w", err)
 	}
+	reverseSwaps = slices.DeleteFunc(reverseSwaps, func(swap *database.ReverseSwap) bool {
+		return !shouldClaimAutomatically(boltz.ReverseSwap, swap.Id, swap.Error)
+	})
+	chainSwaps = slices.DeleteFunc(chainSwaps, func(swap *database.ChainSwap) bool {
+		return !shouldClaimAutomatically(boltz.ChainSwap, swap.Id, swap.Error)
+	})
```
michael1011: When I query the claimable swaps, I expect to get the claimable swaps in return. No implicit post-processing requirement

PR #704, opened 2026-07-13
feat: `scan_to_index` gRPC param for imports

jackstar12:
## Summary by CodeRabbit

* **New Features**
  * Added an optional wallet import scan index for Liquid wallets, persisted as the wallet’s next derivation index.
  * Wallet imports can now scan to a specified derivation index and continue from that point.
  * Added CLI support via `--scan-to-index`.
  * Added configuration support for `liquid-esplora-concurrency` to tune concurrent Liquid wallet syncing.
* **Bug Fixes**
  * Invalid scan-index usage with Bitcoin wallets now returns a clear validation error.  
* **Documentation**
  * Updated RPC API docs and configuration docs to include the new Liquid wallet scan index and concurrency setting.


internal/onchain/liquid-wallet/wallet.go
```diff
@@ -128,6 +129,9 @@ func NewBackend(cfg Config) (*BlockchainBackend, error) {
 		threshold := DefaultMergeThreshold
 		cfg.MergeThreshold = &threshold
 	}
+	if cfg.EsploraConcurrency == 0 {
+		cfg.EsploraConcurrency = 32
```
michael1011: We default to 4 but this says 32. Let's put that in a constant and make it consistent

michael1011 (approved):

PR #712, opened 2026-07-15
feat: limit wallet sync concurrency via semaphore

jackstar12:
## Summary by CodeRabbit

* **New Features**
  * Added a configurable setting to cap concurrent wallet synchronization.
  * Wallet synchronization now runs an initial full scan as part of the normal sync cycle.
* **Bug Fixes**
  * Improved safety when wallets are added or removed during background syncing.
  * Sync work now correctly respects shutdown/cancellation and stops when a wallet is no longer present.
* **Tests**
  * Added coverage to verify the concurrency limit is enforced during wallet syncing.

### DIFF
--- a/internal/onchain/onchain.go
+++ b/internal/onchain/onchain.go
@@ -74,9 +74,9 @@ var DefaultWalletSyncIntervals = map[boltz.Currency]time.Duration{
 	boltz.CurrencyLiquid: time.Minute,
 }
 
-// DefaultMaxSyncConcurrency limits how many wallets sync at the same time,
+// MaxSyncConcurrency limits how many wallets sync at the same time,
 // so a large number of wallets doesn't overload a single backend.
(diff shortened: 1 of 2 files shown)


internal/onchain/onchain.go
```diff
@@ -109,27 +116,91 @@ func (onchain *Onchain) Init() {
 	if onchain.LiquidWalletSyncInterval != 0 {
 		onchain.WalletSyncIntervals[boltz.CurrencyLiquid] = time.Duration(onchain.LiquidWalletSyncInterval) * time.Second
 	}
+	if onchain.MaxSyncConcurrency <= 0 {
+		onchain.MaxSyncConcurrency = DefaultMaxSyncConcurrency
+	}
```
michael1011: This is not actually configurable. Either add it to the config, or make it use the const

PR #725, opened 2026-09-10
fix(cln): log warning and proceed when chanIds are supplied to PayInvoice

### LINKED ISSUE #718: Autoswap reverse swaps always fail on CLN: "chanIds are not supported for cln
### Autoswap reverse swaps always fail on CLN: "chanIds are not supported for cln", channel retried forever

**Environment**

- boltz-client: `v2.12.5-1eab2ad` (docker image)
- Lightning node: `CLN`
- Network: mainnet (BTC -> BTC)

**Summary**

With a CLN backend, every reverse swap created by the lightning autoswap fails
immediately when boltzd tries to pay the invoice:

```
2026-08-01 16:39:59.284 [ERROR] Could not pay invoice lnbc4606530n1p4xug2l...: chanIds are not supported for cln
```

The invoice is truncated for brevity; the full invoice is a standard BOLT11
with nothing sensitive in it. The payment never reaches the network; the swap
is created and then immediately marked `ERROR`. Because the failed swap only
dismisses the channel for
`failure_backoff` (default 3600 s), autoswap recreates the same doomed swap on
every hour. In my case that produced 25 failed swaps in ~25 hours
(2026-07-31 15:39 UTC through 2026-08-01 16:39 UTC), 4 of them ending in
`swap.expired`, the rest stuck in `swap.created`.

**Steps to reproduce**

1. Run boltz-client v2.12.5 with a CLN node and enable lightning autoswap
   (`boltzcli autoswap enable ln`).
2. Have a channel whose inbound balance is below the configured threshold so a
   reverse swap is recommended:

   ```json
   {
     "swap": {
       "amount": "460653",
       "feeEstimate": "2834",
       "type": "REVERSE"
     },
     "channel": {
       "id": { "cln": "956386x1466x0", "lnd": "1051557527738253312" },
       "capacity": "500000",
       "outboundSat": "485653",
       "inboundSat": "14347"
     },
     "thresholds": { "inbound": "150000" }
   }
   ```

3. Wait for the next poll interval. The swap is created, and the invoice
   payment fails deterministically (see log excerpt above).

**Expected behavior**

- The invoice gets paid (either routed through the recommended channel, or
  through any channel with sufficient outbound liquidity), or
- autoswap treats this as a hard failure for the channel instead of retrying a
  swap that can never succeed.

**Actual behavior**

Every autoswap-created reverse swap on CLN fails with
`chanIds are not supported for cln` and the channel is retried every
`failure_backoff` interval forever.

**Workaround (verified on v2.12.5)**

Creating the same reverse swap manually *without* `--chan-id` works. The CLI
leaves `ChanIds` empty in that case, the rejection in `PayInvoice` is never
hit, and CLN routes the payment on its own:

```
docker exec -it boltzd boltzcli createreverseswap btc 460653 <address>
```

Result of the manual swap (`ZRGEcYXRy4au`):

- state `SUCCESSFUL`, status `invoice.settled`
- invoice amount 460,653 sats, received on-chain 458,041 sats
  (boltz fee 2,304 + onchain fee 506)
- claim tx `b9bc2d010801d2b9fca13b74a7b5edd9018e1d428cdbb9992fd1bd5974717376`

One caveat: without chanIds, CLN is free to split the payment across channels
(MPP). The target channel `956386x1466x0` went from 485,653 outbound /
14,347 inbound to 203,416 / 296,583, with the remainder of the invoice routed
through a second channel to the same peer. So the manual swap does fix the
liquidity imbalance, but it cannot force a single channel the way autoswap
intends to.

**Root cause**

Autoswap always pins the payment to the recommended channel by passing its CLN
id as a chanId:

- `internal/autoswap/lightning.go` (`LightningConfig.execute`) builds `chanIds`
  from `recommendation.Channel.Id.Cln` and passes it in
  `CreateReverseSwapRequest.ChanIds` (~lines 401-415).
- The swap stores the chanIds and the nursery pays the invoice with them:
  `internal/nursery/reverse.go` line 69:
  `nursery.lightning.PayInvoice(..., reverseSwap.ChanIds)`.
- The CLN backend hard-rejects any chanIds (`internal/cln/cln.go` lines 226-227):

  ```go
  if len(chanIds) > 0 {
      return nil, fmt.Errorf("chanIds are not supported for cln")
  }
  ```

So on a CLN node every autoswap reverse swap fails deterministically. LND is
unaffected because its `PayInvoice` supports the outgoing channel restriction.

**Suggested fix (one or both)**

1. `internal/cln/cln.go`: don't reject chanIds outright. CLN's `xpay` cannot
   restrict the outgoing channel, so at minimum log a warning and continue with
   normal routing. If channel pinning is a hard requirement, fail at swap
   creation time with a clear error so autoswap can dismiss the channel
   permanently instead of retrying.
2. `internal/autoswap/lightning.go`: only populate `ChanIds` when the backend is
   LND (mirror the SUBMARINE case, where it is already commented out); leave it
   empty for CLN and let CLN route naturally.
3. Consider a longer backoff or permanent dismissal after consecutive failed
   swap attempts on the same channel, otherwise any persistent failure turns
   into an endless retry loop.

**Additional observations**

- No funds are at risk: the payment fails client-side before touching the
  network, and the on-chain side is never funded.
- Around 2026-08-01 09:16-10:36 UTC there was also an ~80 minute
  `lost connection to boltz ws, reconnecting in 15s` loop (websocket close
  1006) that recovered on its own; probably unrelated, but included in case it
  is connected somehow.

**ai help**

Tyagiquamar:
Fixes #718

### Problem
When using Core Lightning (CLN) as the Lightning node backend for `boltz-client`, Lightning autoswap creates reverse swap recommendations containing channel IDs (`reverseSwap.ChanIds`). When `nursery` attempts to pay the invoice for the reverse swap, it passes `chanIds` into `PayInvoice`.

In `internal/cln/cln.go`, `PayInvoice` previously checked:
```go
if len(chanIds) > 0 {
    return nil, fmt.Errorf("chanIds are not supported for cln")
}
```
Because CLN's `Xpay` RPC does not support outgoing channel restriction, returning a hard error caused autoswap-created reverse swaps on CLN nodes to fail immediately.

### Solution
Replace the hard error with a `logger.Warnf` warning log so that `Cln.PayInvoice` logs the unsupported `chanIds` restriction and proceeds to execute `Xpay` to pay the invoice.

### Testing
- Added unit test `TestCln_PayInvoice_ChanIdsWarning` in `internal/cln/cln_test.go` verifying warning fallback and `Xpay` execution.
- Executed `go test -v ./internal/cln/...` in `golang:1.26-bookworm` container:
  ```
  === RUN   TestCln_PayInvoice_ChanIdsWarning
  --- PASS: TestCln_PayInvoice_ChanIdsWarning (0.00s)
  PASS
  ok      github.com/BoltzExchange/boltz-client/v2/internal/cln  0.145s
  ```

## Summary by CodeRabbit

* **Behavior Changes**
  * Payments now proceed through the CLN payment flow when channel IDs are provided.
  * A warning is logged to indicate that the specified channel IDs are ignored.

### DIFF
--- a/internal/cln/cln.go
+++ b/internal/cln/cln.go
@@ -224,7 +224,7 @@ func (c *Cln) PayInvoice(ctx context.Context, invoice string, feeLimit uint, tim
 	retry := uint32(timeoutSeconds)
 
 	if len(chanIds) > 0 {
-		return nil, fmt.Errorf("chanIds are not supported for cln")
+		logger.Warnf("chanIds are not supported for cln; ignoring specified chanIds: %v", chanIds)
 	}
 
 	res, err := c.Client.Xpay(ctx, &protos.XpayRequest{
--- a/internal/cln/cln_test.go
+++ b/internal/cln/cln_test.go
@@ -1,0 +1,44 @@
+package cln
+
+import (
+	"context"
+	"testing"
+
+	"github.com/BoltzExchange/boltz-client/v2/internal/cln/protos"
+	"github.com/BoltzExchange/boltz-client/v2/internal/lightning"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+)
+
+type mockNodeClient struct {
+	protos.NodeClient
+	xpayCalled bool
+}
+
+func (m *mockNodeClient) Xpay(ctx context.Context, in *protos.XpayRequest, opts ...grpc.CallOption) (*protos.XpayResponse, error) {
+	m.xpayCalled = true
+	return &protos.XpayResponse{
+		AmountMsat: &protos.Amount{
+			Msat: 100000,
+		},
+		AmountSentMsat: &protos.Amount{
+			Msat: 101000,
+		},
+	}, nil
+}
+
+func TestCln_PayInvoice_ChanIdsWarning(t *testing.T) {
+	mockClient := &mockNodeClient{}
+	clnNode := &Cln{
+		Client: mockClient,
+	}
+
+	// Passing non-empty chanIds should log a warning and proceed with Xpay rather than returning an error
+	chanIds := []lightning.ChanId{12345}
+	res, err := clnNode.PayInvoice(context.Background(), "lnbc...", 1000, 30, chanIds)
+
+	require.NoError(t, err)
+	require.True(t, mockClient.xpayCalled)
+	require.NotNil(t, res)
+	require.Equal(t, uint(1000), res.FeeMsat)
+}


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

jackstar12:
I assume you have a per-channel strategy configured then? In this case it's expected to have channel id's. If the strategy is set to total channel balance it should work without issues

Tyagiquamar:
Thank you for the clarification, @jackstar12. You are completely right: under a \per-channel\ autoswap strategy, \chanIds\ are required to guarantee that outbound capacity is drawn from that specific channel. Silently ignoring \chanIds\ in CLN \PayInvoice\ violates the per-channel rebalance semantics. Since CLN does not support passing explicit outgoing channel constraints in \PayInvoice\, users on CLN should use the \	otal channel balance\ strategy instead. Closing this PR.

